PR #96, opened 2015-05-01
Provides `PyWrappedTask`

jakirkham:
The class `PyTask` takes some of its arguments through the singular `args` sequence parameter, but gathers its `kwargs` from various keyword arguments that are passed in directly. However, it arbitrarily eliminates some of the parameters like ones beginning with an `_` and, of course, the `func`, `args`, and `stored_data_varname`. This can come as unexpected or surprising to the end user.

To provide a better strategy for wrapping a Python task, we present `PyWrappedTask`, which differs in one crucial way. It has an optional `kwargs` parameter. The `kwargs` parameter takes a dictionary, which is passed to the underlying function untampered. The main benefit of this strategy is that end user can now be confident that all keyword arguments provided this way will be passed untampered (without any removals) to the end function. This way if they require keyword arguments that would clash, they can be guaranteed to have them properly interpreted.

As making this would break the API of `PyTask`, this is presented as a new Task.


computron:
I haven't had a chance to look into this carefully; I understand that PyWrappedTask allows adding underscored vars, but does it take anything away from the original PyTask (i.e. strip features, break backwards compatibility, etc.)? If not, is there any reason to keep around the both PyTask & PyWrappedTask instead of just merging the solution into PyTask?

jakirkham:
Thanks for taking a quick look.

Technically, I believe it would not be backwards compatible with `PyTask`. Hence, a new type. Maybe `PyTask` could be phased out and replaced with `PyWrappedTask`, but I leave that to your discretion.

The cause of the break is that all keyword arguments must be passed as a dictionary to the `kwargs` parameter. So, keywords arguments could not be passed in directly as `PyTask` had.

computron:
Ok - I think my preference is to try to have one solution if at all possible, especially since the problem statement doesn't seem like there are two very separate use cases that require separate solutions.

I will take a look later this weekend to try and understand the differences between the classes, as well as try to figure out a good way forward. I will hold off on merging until deciding whether it makes sense to have two classes. Hopefully this doesn't interfere with your work (as I assume you are working off your own fork), but let me know if it does.

jakirkham:
That makes sense and I agree that having two very similar things as a long term solution is bad. However, I think moving towards the API of `PyWrappedTask` is simpler and more predictable. So, if if there is another way that you have in mind of getting there, I am interested in hearing your thoughts.

> Hopefully this doesn't interfere with your work...

Not a problem. I have my own copy until we find a satisfactory solution.

jakirkham:
Thanks for finding a way to make this work.

Issue #98, opened 2015-05-05
Running `rapidfire` in Python results in multiply nested directories 


jakirkham:
It isn't obvious to me from the spec that this is intended. I would expect that all directories would be at the same level as if one had run `launch_rocket` multiple times. Instead, it appears that the directories become deeply nested. This seems to be caused by multiple calls to `os.chdir`. If this is indeed intended and expected behavior, I would expect at the minimum that we would change back to the original directory after a run.

PR #99, opened 2015-05-05
Rapidfire switch back to cwd at the end.

jakirkham:
Fixes materialsproject/fireworks#98. I have verified that the tests fail when my solution is not included. Also, this works when testing in workflows.


jakirkham:
One open question that I am interested in feedback on is the following, if `m_dir` is set, should we end in `m_dir` or should we end up the directory we were in before going to `m_dir`.

jakirkham:
Toggling open to force Travis CI to rerun due to spurious failure see issue ( https://github.com/materialsproject/fireworks/issues/100 ). Also, note that it had passed previously in this build ( https://travis-ci.org/jakirkham/fireworks/builds/61310365 ).

computron:
>> One open question that I am interested in feedback
We probably want to end up in m_dir

jakirkham:
Thanks for the clarification. Then it is ready to go.

Issue #100, opened 2015-05-05
Frequent spurious failures from  `test_rerun_timed_fws`


jakirkham:
I have seen this test fail often and suspect there is something wrong with the test itself (particularly as it is a timed test). Here is an example of it failing when I had only made changes to docs and docstrings ( https://travis-ci.org/materialsproject/fireworks/builds/60772474 ).

    ======================================================================
    FAIL: test_rerun_timed_fws (fireworks.core.tests.test_launchpad.WorkflowFireworkStatesTest)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/home/travis/build/materialsproject/fireworks/fireworks/core/tests/test_launchpad.py", line 710, in test_rerun_timed_fws
        self.assertEqual(fw_state, fw_cache_state)
    nose.proxy.AssertionError: 'RUNNING' != 'READY'
    - RUNNING
    + READY

computron:
Ok, I am assigning this to @mbkumar who wrote and is in charge of this test

jakirkham:
Glad to hear someone will take a closer look. Thanks.

PR #101, opened 2015-05-07
fix for FW files and log

gpetretto:


jakirkham:
FWIW, it seems the failure is spurious and is the same one mentioned here ( https://github.com/materialsproject/fireworks/issues/100 ).

Issue #93, opened 2015-04-30
Calling `lpad webgui` opens two identical tabs


jakirkham:
As the title says, running `lpad webgui` seems to open the same page in two different tabs or windows.

computron:
Hi,

I can't reproduce this problem on a Mac OS/X Yosemite; calling "lpad webgui" opens up only a single tab. Would you be able to provide some details of your environment?

Note that it's unlikely I will be able to debug this issue since I can't reproduce, and nothing in the code looks obviously off. If you have a solution I'm happy to merge it; one workaround (other than closing the added tab) is to run the command in server mode ("lpad webgui -s") and open up the browser yourself.

jakirkham:
Alright, let me look into this further. Any pointers as to place to look?

computron:
Ok, the browser opening is in: fireworks/scripts/lpad_run.py:411

```
def webgui(args):
    os.environ["FWDB_CONFIG"] = json.dumps(get_lp(args).to_dict())
    from fireworks.flask_site.app import app
    from multiprocessing import Process
    p1 = Process(target=app.run, kwargs={"host": args.host, "port": args.port})
    p1.start()
    if not args.server_mode:
        import webbrowser
        time.sleep(2)
        webbrowser.open("http://{}:{}".format(args.host, args.port))
    p1.join()
```

jakirkham:
Perfect, thanks. I will take a look.

computron:
I am closing this issue for now - please re-open if needed

jakirkham:
Yeah, it appears I am not allowed to reopen issues on your issue tracker. Regardless, this appears to be a bug in the browser I am using.

jakirkham:
Actually, I disagree with that assessment now. I tried a variety of browsers only Safari did not have this problem. This is still a bug.

jakirkham:
@computron please try using Firefox or Chrome and let me know if you still are not able to reproduce this.

gpetretto:
Hi,
I could reproduce the problem on my laptop (linux+chrome/ff). 
Replacing the line
    p1 = Process(target=app.run, kwargs={"host": args.host, "port": args.port})
with 
    p1 = Process(target=app.run, kwargs={"host": args.host, "port": args.port, "debug": False})
in the webgui function seems to fix the problem. Can you check if this is the case for you as well? 
Thanks

jakirkham:
[...]

PR #105, opened 2015-05-13
fix for issue #93: Calling `lpad webgui` opens two identical tabs

gpetretto:


jakirkham:
Thanks for putting this together.

PR #108, opened 2015-06-19
Add Flask-paginate as a requirement

jakirkham:
It appears that `Flask-paginate` is needed to get the Web GUI running so added it as a requirement.


jakirkham:
Note the error here is due to the know test failure ( materialsproject/fireworks#100 ).

PR #109, opened 2015-06-19
Restore workaround for issue materialsproject/fireworks#100

jakirkham:
As mentioned in issue ( materialsproject/fireworks#100 ),the test, `test_rerun_timed_fws`, still has spurious failures as evidenced by the most recent builds on master ( https://travis-ci.org/materialsproject/fireworks/builds/65300009 ). So, bringing back the workaround until the problem is better understood and the necessary fix applied.


jakirkham:
@computron, can we get this test skip back in?

jakirkham:
Thanks.

PR #113, opened 2015-07-13
add support for IBM LoadSharingFacility to common_adapter.py

zulissi:
common_adapter.py currently supports IBM LoadLeveler, which has been replaced with IBM LoadSharingFacility (LSF).  

Adding support for LSF was pretty trivial; the only problem I found was that the job submission utility only processed the file headers (queue name, job name, etc) correctly when the file was read in through stdin (i.e. 'bsub < submissionfile').  This may be a problem with our local cluster installation; we currently use a custom wrapper that reads the file ahead of time and converts the header into command line arguments.  I don't think the proposed special-case for the submission (lines 140-145) would break support even if submitting as a file worked correctly on someone else's cluster.

### DIFF
--- a/fireworks/user_objects/queue_adapters/LoadSharingFacility_template.txt
+++ b/fireworks/user_objects/queue_adapters/LoadSharingFacility_template.txt
@@ -1,0 +1,13 @@
+#BSUB -n $${ntasks}
+#BSUB -W $${walltime}
+#BSUB -q $${queue}
+#BSUB -J $${job_name}
+#BSUB -o $${job_name}-%J.out
+#BSUB -e $${job_name}-%J.error
+
+$${pre_rocket}
+cd $${launch_dir}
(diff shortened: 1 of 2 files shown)


computron:
Regarding this line:

inputFile=open(script_file,'r')

It looks like you need to close() the input file. If there is a way to use the "with open()" syntax, that would be better.

zulissi:
Sure; changed it to the 'with open()' syntax.

computron:
Thanks! It's merged

jsmedmar:
Wondering if there is any documentation instructing how to use Fireworks with LSF?

computron:
Not that I know of, just the basic queue submission instructions that are applicable to many systems.

Note - the official help list is here:
https://groups.google.com/forum/#!forum/fireworkflows

PR #123, opened 2015-10-01
Cast query arg to lpad report from str to dict

dwinston:
I'd like to run reports like
```bash
$ lpad report -i months -q '{"spec.mpsnl.about.remarks": "MP user submission"}'
```
but it seems the query arg is left as a `str`, so I get an `AttributeError: 'str' object has no attribute 'update'` [here](https://github.com/materialsproject/fireworks/blob/f112342dd26f06d5b4622f117ca2cc79679e7095/fireworks/features/fw_report.py#L52).

My PR fixes this issue for me so that I can run the above line with success.

### DIFF
--- a/fireworks/features/fw_report.py
+++ b/fireworks/features/fw_report.py
@@ -1,5 +1,6 @@
 from __future__ import division
 
+import ast
 from collections import OrderedDict
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


computron:
Thanks! I should have probably tested this function, but glad that you caught it

PR #124, opened 2015-10-12
ENH: Added methods to merge workflows together

henrus:
Hi,
this pull request provides methods to merge workflows together. I needed this functionality for the MoDeNa project (https://github.com/MoDeNa-EUProject/MoDeNa) because I would like to build the workflows step by step. Please let me know whether this is acceptable - I am happy to apply changes or rethink the design.
Best Regards,
Henrik

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -861,6 +861,54 @@ class Workflow(FWSerializable):
 
         return updated_ids
 
+    def addAfterAll(self, wf):
+        """
+        Adds a workflow such that it is executed after the current one.
+        """
+
+        root_ids = wf.root_fw_ids
+        leaf_ids = wf.leaf_fw_ids
+
+        my_leaf_ids = self.leaf_fw_ids
+
+        for new_fw in wf.fws:
+            if new_fw.fw_id > 0:
+                raise ValueError(
+                    'FireWorks to add must use a negative fw_id! Got fw_id: '
+                    '{}'.format(
+                        new_fw.fw_id))
+
+            self.id_fw[new_fw.fw_id] = new_fw  # add new_fw to id_fw
+
+            if new_fw.fw_id in leaf_ids:
+                self.links[new_fw.fw_id] = []
+            else:
+                self.links[new_fw.fw_id] = wf.links[new_fw.fw_id]
+
+        for fw_id in my_leaf_ids:
+            self.links[fw_id].extend(root_ids)  # add the root id as my child
+
+    def addNoLink(self, wf):
+        """
+        Adds a workflow without connecting it.
+        """
+
+        leaf_ids = wf.leaf_fw_ids
+
+        for new_fw in wf.fws:
+            if new_fw.fw_id > 0:
+                raise ValueError(
+                    'FireWorks to add must use a negative fw_id! Got fw_id: '
+                    '{}'.format(
+                        new_fw.fw_id))
+
+            self.id_fw[new_fw.fw_id] = new_fw  # add new_fw to id_fw
+
+            if new_fw.fw_id in leaf_ids:
+                self.links[new_fw.fw_id] = []
+            else:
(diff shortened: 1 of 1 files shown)


computron:
Hi Henrik,

Thanks very much for making this change. You are correct that adding workflows together is a needed functionality. I had two questions regarding your implementation:

1. What is the use case of adding a workflow without connecting it to the previous one (addNoLink)? (as opposed to simply adding a new independent workflow). As far as I understand the two workflows would run independently if there is no link between them. But, I guess you could archive the workflows together or get unified information as a kind of a "hack" to group together certain workflows. Is this the intention?

2. Rather than two methods, my suggestion is to have one method with an optional parameter for fw_ids that get connected to the root workflow. e.g.:

def append_workflow(self, wf, parent_fw_ids):

If the parent_fw_ids is set to None, then it is the equivalent of "no links" (provided that this use case makes sense)
If the parent_fw_ids is set to single number or array of numbers, then those become the parents.
If the parent_fw_ids is set to "leaves", then it will attach the workflow to all the leaves of the parent WF (equivalent to the current method).

What do you think? I think it would be a straightforward change to your current implementation and reuse most of the code from the two methods into one.

computron:
Hi Henrik,

I am merging this request, and will make a few changes as per the comment above. Thanks again for your contribution!

computron:
Actually, I just realized that the added code probably be achieved with the existing code.

i.e., a
-addAfterAll is the same as the existing add_wf_to_fws function with fw_ids set to the wf.leaf_ids, and other parameters set to False
-addNoLink is the same as the existing add_wf_to_fws function with fw_ids set to [] and other parameters set to False

Is there a difference in functionality that I am leaving out?

henrus:
Hi,

in fact add_wf_to_fws is where I started from, but I cannot remember why I had to change it. Anyway, I tried what you suggested

    def addAfterAll(self, wf):
        self.add_wf_to_fws(wf, wf.leaf_fw_ids, pull_spec_mods=False)

    def addNoLink(self, wf):
        self.add_wf_to_fws(wf, [], pull_spec_mods=False)

This results in 'refresh' throwing an error.

zip --- 

  File "./initModels_rh", line 60, in <module>
    initWfs.addNoLink(m.initialisationStrategy().workflow(m))
  File "/home/henrus/lib/python2.7/site-packages/modena/Strategy.py", line 124, in addNoLink
    self.add_wf_to_fws(wf, [])
  File "/usr/local/lib/python2.7/dist-packages/fireworks/core/firework.py", line 874, in add_wf_to_fws
    updated_ids = self.refresh(new_fw.fw_id, set(updated_ids))
  File "/usr/local/lib/python2.7/dist-packages/fireworks/core/firework.py", line 939, in refresh
    for child_id in self.links[fw_id]:
KeyError: -4

zip ---

Investigating ...

computron:
Ok

Note that I have already modified the latest FWS to allow for these functions when using the appropriate arguments. i.e., see Workflow.append_wf()

-addAfterAll is the same as the existing function with fw_ids set to the wf.leaf_ids set to False
-addNoLink is the same as the existing function with fw_ids set to [] 

If you want to append to a Workflow that is already in the Launchpad, see LaunchPad.append_wf()

henrus:
Ok. Didn't see it. I will try your implementation in second.

However, I am pretty sure that they are different

addAfterAll/addNoLink:
            if new_fw.fw_id in leaf_ids:
                self.links[new_fw.fw_id] = []
            else:
                self.links[new_fw.fw_id] = wf.links[new_fw.fw_id]

add_wf_to_fws(self, new_wf, fw_ids, pull_spec_mods=True, detour=False):

                if new_fw.fw_id in leaf_ids:
                    # detour=False -> code stripped
                    self.links[new_fw.fw_id] = []
                else:
                    self.links[new_fw.fw_id] = new_wf.links[new_fw.fw_id]

(the last line is different).

henrus:
Forget my last message. All clear and working now.

PR #125, opened 2015-10-15
update wflock and handling of inconsistencies in the db

computron:


computron:
Hi Guido,

Thanks for providing these changes. I have reviewed them and think they are good. I would make the following suggestions (either for you to implement, or for me to implement, either is fine just let me know):

* in the detect_lostruns function (and subsequently "lpad detect_lostruns" script) there is the new options of checking inconsistency and refreshing workflows where detected. I would remove these as options and just have them be True. i.e., always check for inconsistencies and refresh inside of detect_lostruns. I don't think that adding this feature will signficantly reduce performance or cause problems (since you are just refreshing workflows), so it is better to just apply the fix rather than have the user decide.

* I had a minor comment regarding the "final_state" variable in rocket.run() - simply, to reuse this variable when calling complete_launch() rather than listing the same string twice.

i.e. instead of:

    final_state = 'FIZZLED'
    lp.complete_launch(launch_id, m_action, 'FIZZLED')

to have:

    final_state = 'FIZZLED'
    lp.complete_launch(launch_id, m_action, final_state)

* Finally, it looks like the Travis CI build succeeds for Python 2.7 but fails for Python 3.3. Do you have any idea what might be the problem? The error looks like this:

```
ERROR: test_fix_db_incosistencies_completed (fireworks.core.tests.test_launchpad.WFLockTest)
----------------------------------------------------------------------
  File "/home/travis/build/materialsproject/fireworks/fireworks/core/tests/test_launchpad.py", line 902, in setUp
    fireworks.core.launchpad.WFLock.__init__.im_func.func_defaults = (3, False)
nose.proxy.AttributeError: 'function' object has no attribute 'im_func'
```

Thanks again for making these important changes.

gpetretto:
Hi Anubhav,

I fixed the test, there were actually a couple of things that weren't python 3 compatible.

For the first point, the only problem I see is that right now, unless explicitly requested by the user, detect_lostruns does not perform any action on the db. Making automatically refresh the db in case of incompatibilities could be unexpected for an old user. In case you prefer to remove the option, I'd suggest to mention this change in the changelog.

computron:
Hi Guido,

Great, thanks!

Regarding your point on detect_lostruns, perhaps we can then keep the "refresh" option but remove the "inconsistent" option. i.e., it *always* checks for inconsistent runs, but only refreshes the runs if the user requests it. What do think?

gpetretto:
This seems like a good compromise. 
Do I have to add some documentation?

computron:
You can add the level of documentation you feel is appropriate to the failures tutorial (if desired). I will supplement whatever you write before releasing the next version.

gpetretto:
I added a short description of the problem and how to fix it in the tutorial. I hope it is fine.

PR #129, opened 2015-10-29
fix to fw_serializers to fix dict/list issue

montoyjh:
Using "obj.__class__ is" rather than "isinstance" seems like it might not be the best way to fix this issue, but seems to work.

### DIFF
--- a/fireworks/utilities/fw_serializers.py
+++ b/fireworks/utilities/fw_serializers.py
@@ -98,7 +98,7 @@ def _recursive_load(obj):
     if hasattr(obj, '_fw_name'):
         return obj
 
-    if isinstance(obj, dict):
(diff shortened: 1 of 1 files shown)


computron:
Hi Joey,

Thanks for all your detective work on this issue. I fixed this in a different way to prevent MSONable objects that are already deserialized from being deserialized again. Can you check whether the newest code in Github works? I confirmed that in the old code, deserializing and Incar.as_dict() two times in a row would result in a dict object, whereas in the new code, two consecutive deserializations maintains an Incar object.

montoyjh:
Yeah, this makes more sense.  Tested it with the dynamic workflow I was having trouble with earlier, and it seems to check out.

PR #132, opened 2015-11-08
Cytoscape workflow drawing, with dagre layout, and handlebars templated node details

lordzappo:
Hi Fireworks Friends,

This commit should cleanly encapsulate workflow drawing  on the wf_details.html page.  
* cytoscape.js draws a canvas object that is performant even for very large workflows
* dagre.js will automatically suggest a layout
* cytoscape-panzoom adds googlemaps like controls
* Handlebars.js is used to template node details responses
* Two json routes are added to the flask app for cytoscape and handlebars to suck data from via AJAX requests after page load

Known issues:
* The cytoscape panzoom css is a little janky looking after I merged it with your css - you'll probably want to touch that up. (It has its own css you can edit)
* The node details that are exposed via the handlebars template in wf_details.html are not the most helpful ones, you should tweak them - we use different ones in our version of fireworks, but they are tied to our filesystem so aren't generalizable for all of your systems.
* It takes about a megabyte of javascript to drive the graph.  Moving cytoscape/dagre to bottom of page to load later and putting a little bit of loading icon magic in might make a better user experience.  I can help with this later if this isn't clear.
* the colors of the cytoscape drawing have to be defined in the Flask app according to my current understanding of cytoscape.js- if it were SVG based we could use css classes, but cytoscape no longer is for performance reasons.


computron:
Thanks so much!

I am going to merge the commits now and work on integration over the next week. I might have a few questions during the merging process but hopefully most of it will go smoothly.

Thanks again for contributing -

Anubhav

lordzappo:
Hi again,

I will contribute the more information dense front page whenever you like, I wanted to separate the commits.  It will use bootstrap 3 to get the progress bars.  Bootstrap 3 would also give you the "panel-heading", "panel-body" classes I am calling in the step details handlebar template in wf_details.html. 

It will also affect home.html as the grid layout system class names changed between 2 and 3. I have the translation, but what I'm saying is it will potentially affect many files, albeit in a relatively superficial way.

CCH

PR #136, opened 2015-11-23
few bugfix and updates

gpetretto:
- fizzled jobs require a completion time for offline runs
- fix pagination on the webgui
- added a link to the fws in the wf_details page


computron:
Thanks!

PR #138, opened 2015-12-04
ENH: Reduce complexity of append_wf by moving check out of loop

henrus:
In my opinion there are two issues:
a) If the check fails it potentially leaves an half-finished Workflow for somebody in case somebody is trying to recover from the ValueError.
b) The check is executed len(new_wf.fws) time which is unnecessary.
Henrik


henrus:
In addition the patch fixes the the following bug for
wf = Workflow([])
wf.append_wf(Workflow([t1, t2], [])
Fails

computron:
Thanks!

PR #144, opened 2016-03-22
small improvements for offline mode

gpetretto:
- only launches for worker specified by -w will be recovered (one needs to run the command on the actual worker anyway)
- recover_offline can optionally print the exceptions
- fix: recover_offline sets the state of the FW to running when needed
- tests


computron:
Thanks!

PR #147, opened 2016-05-02
FileTransferTask bugfixes, added user parameter

dotsdl:
Fixes [a misuse of `get`](https://github.com/materialsproject/fireworks/blob/master/fireworks/user_objects/firetasks/fileio_tasks.py#L98) in the `FileTransferTask` relevant for remote transfers. Also ensures that if `src` is a file, `dest` is still interpreted as a directory (as it should according to the class docstring).

Additionally, adds a `user` parameter for specifying the username to authenticate against on the remote server.

If this is too many things in one PR, I'm happy to split them up.


computron:
Thanks very much!

PR #150, opened 2016-05-06
Can now make a fresh copy of an existing Workflow.

dotsdl:
There are occasions in which a new Workflow needs to be created from an
existing one, but with updated Firework ids to ensure no collisions with
any Workflows it is appended to. Such collisions can occur when a
Workflow has been serialized and stored within a Firework spec, then
appended by that Firework to the Workflow it is a part of (as an ``addition``).
The result of these collisions are often malformed links.

Building a fresh Workflow from that stored avoids this problem, since
the member Fireworks are given fresh ids that have not been used within
the current session.

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -30,7 +30,7 @@ Zach Ulissi added support for IBM LoadSharing facility.
 
 William Scullin added ALCF Cobalt support and helped stomp bugs.
 
-Brian Foster, kpoman, jakirkham, Ganesh Panchapakesan, Patrick Huck, Donny Winston, Joey Montoya, Henrik Rusche, David Cossey, David Doston, specter119, and Felix Zapata helped stomp bugs and provide improvements.
+Brian Foster, kpoman, jakirkham, Ganesh Panchapakesan, Patrick Huck, Donny Winston, Joey Montoya, Henrik Rusche, David Cossey, David Dotson, specter119, and Felix Zapata helped stomp bugs and provide improvements.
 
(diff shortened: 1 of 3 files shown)


computron:
Thanks! A few suggestions

1. To copy, note that copy.deepcopy() can actually be finicky. I would suggest doing Workflow.from_dict(wflow.to_dict()). This serialization is guaranteed to be correct (if it isn't you will have other problems). If you think this is non-intuitive to most folks, I would then override the __copy__() and __deepcopy__() methods of Workflow so that anyone that tries to use the "standard" copy methods will get the correct result.

2. I would separate out the reset function of the copied workflow, i.e. add:

Workflow.reset(reset_ids=False) - this will reset the states and fw_states dict to WAITING. If the reset_ids is set to True, then it will also reset the ids.

3. Then Workflow.from_wflow() is just two lines of code - one that copies the old one, and one that resets the new one

dotsdl:
Pushed changes reflecting your feedback. It definitely looks a lot better now. Is there a good place to add tests for these methods to the test suite?

computron:
Thanks!

Tests can go in fireworks/core/tests/test_firework.py

Note that there is at least one bug in that the word "reset_ids" is spelled "rest_ids" in 2 places (once in the code itself, once in docs)

dotsdl:
Thanks for catching that. Fixed! Also added a simple test for checking that the copied workflow looks the same in terms of its dependency links. Let me know if you want something more thorough.

computron:
Thanks!

PR #151, opened 2016-05-06
Feature filetransfer retry

dotsdl:
Added ``retry`` option to ``FileTransferTask``. It's a bit crude in that it just runs ``self.run_task`` again if the `try` block fails and the ``retry`` option is ``True``. As written it will keep trying until the recursion limit is reached.

Happy to adjust the way this is implemented to make it less crude. At the moment this is a (useful) hack I use to guard against remote transfer failures, which are frequent enough (for me) as to make this a must-have.

### DIFF
--- a/fireworks/user_objects/firetasks/fileio_tasks.py
+++ b/fireworks/user_objects/firetasks/fileio_tasks.py
@@ -71,6 +71,8 @@ class FileTransferTask(FireTaskBase):
         - server: (str) server host for remote transfer
         - user: (str) user to authenticate with on remote server
         - key_filename: (str) optional SSH key location for remote transfer
+        - max_retry: (int) number of times to retry failed transfers; defaults to `0` (no retries)
+        - retry_delay: (int) number of seconds to wait between retries; defaults to `10`
     """
     _fw_name = 'FileTransferTask'
     required_params = ["mode", "files"]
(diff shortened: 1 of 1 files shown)


computron:
Hi David,

How about:

1. Rename "retry" to "max_retry" and have it be an integer. 0 indicates no retry.
2. Add a little delay before retrying, i.e. time.sleep(), just in case the cause of the error was just some temporary thing
3. When calling run_task again, change the max_retry to be max_retry - 1. That will prevent infinite recursion (which will might kind of hammer the OS especially when there is no sleep time)

dotsdl:
@computron these are definitely a better way to go about this; I've pushed changes accordingly. Is there a good place for adding in tests for these options?

computron:
Hi David,

Thanks for making these changes.

The tests can go in fireworks/user_objects/firetasks/tests/test_fileio_tasks.py

Note that there is a bug in this line:
retry_delay = self.get('max_retry', 10)

It should probably read:
retry_delay = self.get('retry_delay', 10)

dotsdl:
Not sure the best way to test this, since it's a bit of a random failure that it seeks to address. Happy to run with suggestions, though. Thanks for catching that typo; fixed!

computron:
(note - there was also a missing import for the time module, but I pushed a patch for that)

dotsdl:
Thanks!

PR #152, opened 2016-05-06
Always unreserve if something went wrong

gpetretto:
It should be better to unreserve the FW at the end of the queue launch process.


computron:
Hi Guido,

Thanks! But the unittests fail (see travis-ci message above)

I think probably you just need to update the monty version dependency in requirements.txt and setup.py. Could you try that and see if it fixes it?

gpetretto:
Sorry, I didn't notice that the monty version in the requirements was an old one. I set it to the latest version.

PR #153, opened 2016-05-06
Fixed problem with ssl_ca_file ... if it is None, just dont pass it t…

davidwaroquiers:
…o MongoClient.

For some reason, this did not work ... maybe due to (old) version of MongoDB at our site (2.4) or version of pymongo (3.2.2) or combination of the two ? Anyway this fixes the problem without removing the ssl_ca feature.


computron:
Ok - can you just add a comment to the code as to why there is the "if" statement? Otherwise the next person that comes along might want to "fix" it by just making it a single statement again.

computron:
Sorry - somehow I did not keep of track of this merge request. I am pulling it now.

PR #155, opened 2016-05-24
Fix SGE qlaunch rapidfire --maxjobs_queue detection

dotsdl:
Using ``qlaunch rapidfire --maxjobs_queue <NUM>`` for launching rockets to an SGE queue did not properly detect the number of jobs in the queue. This is because by default the ``qstat`` output doesn't include the queue assignment for jobs that are not yet running, so in the case where ``<NUM>`` is greater than the number of jobs actually running, jobs will continue to be submitted (especially for ``--nlaunches infinite``) well beyond ``<NUM>``.

This PR adds a special case (as already exist for some other resource managers) for SGE, changing the ``qstat`` call used by the qlauncher to specify the queue name *and* the user, then counting the lines present. 

SGE does not leave completed jobs in the ``qstat`` output to my knowledge (doesn't happen on our installation, but if anyone knows otherwise we can add exclusions in the parsing), so we just do simple line counts of the resulting output from the ``qstat`` call.

### DIFF
--- a/fireworks/user_objects/queue_adapters/common_adapter.py
+++ b/fireworks/user_objects/queue_adapters/common_adapter.py
@@ -136,9 +136,9 @@ class CommonAdapter(QueueAdapterBase):
             #Just count the number of lines
             return len(output_str.split('\n'))
         if self.q_type == "SGE":
-            # output string has states and separator as first two lines
-            # one more due to trailing '\n' gets the right number
-            return len(output_str.split('\n')) - 3
+            # want only lines that include username;
+            # this will exclude e.g. header lines
+            return len([l for l in output_str.split('\n') if username in l])
 
         count = 0
         for l in output_str.split('\n'):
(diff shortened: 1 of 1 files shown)


dotsdl:
Finger slipped; still editing PR. Sorry. :/

computron:
Hi David

Thanks! It looks like the unit tests break - can you take a look?

In particular, make sure that your method for parsing njobs from SGE is robust for different SGE environments. You might, for example, count only lines that contain a certain keyword (e.g. like the username or some other universal indicator of it being a job line rather than header line) rather than arbitrarily subtract 3.

dotsdl:
@computron agreed, the parsing of the ``qstat`` output is too fragile here. Pushed a fix according to your advice, counting only lines with ``username`` present. This should be robust.

dotsdl:
Waiting for Travis to finish to see if failure persists; had a hard time grokking what was happening on the previous failed build.

dotsdl:
Okay...perhaps the failure before was a hiccup on Travis' part? Weird...

computron:
Hi David,

Thanks!

Note the failure before was not a hiccup, it was actually due to your change in the way the SGE jobs were counted which broke one of the tests. If you click the red 'X' it will show you exactly what test failed (it was the SGE counting test). The new code passes the test.

dotsdl:
@computron ah now I see it:

```
  File "/home/travis/build/materialsproject/fireworks/fireworks/user_objects/queue_adapters/tests/test_common_adapter.py", line 113, in test_parse_njobs
    self.assertEqual(p._parse_njobs(sge, "ongsp"), 3)
AssertionError: 4 != 3
```

I should have looked more closely. Glad the fix still satisfies the test.

PR #158, opened 2016-06-02
master

xhqu1981:
replaced two deprecated pymongo method: insert() and find_and_modify(). Tested with test_launchpad, no further test.


computron:
Thanks!

PR #162, opened 2016-06-15
Fixing Documentation - Introduction

ardunn:
Documentation > Introduction >  optional python examples >  rapidfire example
This example won't work unless FWorker is imported from fireworks.

thx to @saurabh02


computron:
Thanks!

PR #164, opened 2016-06-28
Add Options to Leave Head Node Out in Job Packing

xhqu1981:
An option to exclude head node from compute node list is added to mlaunch. As MOM node is no longer available on Edison/Cori, heave jobs, e.g. VASP will also run on the head node. this option will help to reduce the presure on head node and therefore improve computational efficiecy if many nodes are used. By default, head node is not excluded. i.e. keep the node list as is.


computron:
thanks!

PR #165, opened 2016-07-06
Enhance webgui server mode using gunicorn

dwinston:
Use the [Gunicorn](http://gunicorn.org/) WSGI server for `lpad webgui` in server (`-s`) mode. Gunicorn is a proven technology for running Flask apps in production, and is only a `pip install` away.

Also in server mode, re-create the `flask_site.app.lp` LaunchPad after process forking to avoid MongoDB client deadlock. See http://api.mongodb.com/python/current/faq.html#using-pymongo-with-multiprocessing.

This repo branch is currently powering http://fireworks.dash.materialsproject.org/. While running the Flask app directly is fine for local job monitoring, I found it to crash periodically in production. Thus, I propose this as a transparent (i.e., one still simply runs `lpad webgui -s ...`) solution.


computron:
Thanks!

PR #178, opened 2016-09-16
Recovery fix

montoyjh:
This PR attempts to fix two issues in rerunning fireworks at the task-level.

* Upon recoverable exits for fireworks, will store the accumulated stored_data, mod_spec, and update_spec in the "recovery" field of stored data in the action of the failed launch.
* Upon rerunning fireworks at the task level, fireworks will update all_stored_data, all_mod_spec, and all_update_spec with the contents of the corresponding fields in the failed launch.
* `lpad rerun_fws --clear-recovery` will unset the _recover_launch key in the firework spec.  This maintains the current operation of rerun_fws, in which fireworks rerun after being rerun once at the task level will continue to be rerun that way, but allows users to rerun from the beginning if they want to clear the recovery data.  Note that this option will not unset the "_launch_dir" key, even if that key was set in the operation of `lpad rerun_fws --task-level --previous-dir`.

I've tested this with my use case, in which a certain mod_spec (specifically calc_locs from matmethods) set in a firetask prior to the failed firetask was not recovered in the task-level rerun, and thus did not pass that modification to children fireworks, and it is resolved.  This also adds a bit more complexity to the task-level rerunning scheme, unfortunately.  There may be a way to more simply implement this in a larger refactor of the way fireworks are rerun, but this is my most straightforward short term solution.


computron:
Thanks! I think the task level still requires a bit of work, but this at least improves the situation.

Can you add docs about the command to the tutorial files? in docs/rerun_tutorial.rst

montoyjh:
Just added the docs.

computron:
Thanks!

PR #192, opened 2016-11-14
When using the _pass_job_info_feature in several fireworks following …

davidwaroquiers:
…each other, job_info was duplicated. Now the info about the current fw is added to the list only if it is not yet in there.


computron:
Thanks! Sorry for the delay on this

PR #193, opened 2016-12-07
Email report

aykol:
lpad option to send the report as an email using the mail command.

### DIFF
--- a/fireworks/scripts/lpad_run.py
+++ b/fireworks/scripts/lpad_run.py
@@ -542,9 +542,9 @@ def report(args):
     title_dec = "-" * len(title_str)
     if email:
         try:
-            os.system('echo "{}"| mail -s {} {}'.format([title_dec, title_str, title_dec,
-                                                         fwr.get_stats_str(stats)].join('\n'),
-                                                         "Fireworks Report", args.email))
+            os.system('echo "{}"| mail -s {} {}'.format('\n'.join([title_dec, title_str, title_dec,
+                                                         fwr.get_stats_str(stats)]),
+                                                         '"Fireworks Report"', args.email))
             print("Report sent to {}.".format(email))
         except:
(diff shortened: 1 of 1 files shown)


shyuep:
I strongly discourage this. The approach assumes that the system has unix mail commadn. That is BAD design. If you want to do this, use the proper python email package. Either than, just run the lpad pipe to actual mail on your system adn forget about adding this to Fireworks. That's what Anubhav and I use for monitoring our work.

computron:
I just talked to Murat about this. I think it will be nice to help steer people about how to send the e-mail report. Yes, if you know about Unix mail and can put two and two together, you will figure out that you can send an e-mail report with FireWorks. But sometimes people need help putting two and two together.

Rather than add extra code to FireWorks, I would suggest just appending the FireWorks documentation with an example of how to use Unix mail to send reports. Then people know how to do it and this also steers people towards the use case. The syntax is about as simple as using the "--email" option within FireWorks so I also don't see much point in adding something new to FireWorks.

shyuep:
Ok. Yes, the # of characters typed is the same. But even if such an option is desired, I would recommend using an non-OS specific way to do this, see https://docs.python.org/2/library/email-examples.html. That requires more code than relying on Unix mail command, but should work on any platform.

computron:
But then you need to specify the SMTP server, which I think is beyond what most casual users will be capable of setting up. At least the default mail will just let you set the recipient and away we go

shyuep:
Well, you also need to set up SMTP on Linux. The only difference is that on Linux, this is usually set up by default. If you want to do this on Mac or Windows, you have to set up SMTP, AND use the non-OS specific way to send mail. 

Anyway, I am not insisting that this feature be implemented. But if you want to implement it, that would be the way to do it.

PR #195, opened 2016-12-08
Pause fw

shyamd:
Added in pause_fw to documentation
Added in pause_fw to unit tests
Added in PAUSED state to fireworks
Added in pause_fw to LaunchPad
Added in pause_fw into lpad
Added in PAUSED state into webgui

### DIFF
--- a/docs/defuse_tutorial.rst
+++ b/docs/defuse_tutorial.rst
@@ -29,7 +29,7 @@ where ``<FW_IDS>`` is a numerical id of one of the FireWorks in the workflow you
 Canceling and restarting individual FireWorks
 =============================================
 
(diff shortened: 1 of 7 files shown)


computron:
If I call "lpad pause_fws" on a bunch of Fireworks, but one of them is not pausable, will it choke the remaining ones due to the ValueError that is thrown? Maybe you want to catch that error and simply report it rather than quit.

For the docs it would be nice to explain that pause_fws and defuse_fws are essentially the same thing apart from nomenclature, and so the pause function is mainly to help you organize runs. The only difference is that defuse_fws will also work on RUNNING fireworks - partly because this is needed by the detect_lostruns command to mark those "lost" RUNNING fireworks as defused.

computron:
Thanks!

PR #197, opened 2016-12-08
filepad

matk86:
Filepad: interface to dump/retrieve the contents of arbitrary files to gridfs and attach a label to it.

* uses the same database as that of launchpad
* supports add, delete, update operations by label(also by id and by general query)
* fireatasks for adding and deleting files to the filepad
*  unittests for the filepad api and the firetasks


computron:
Thanks! Can you add docs? Otherwise no one will know how to use this

matk86:
i have added docs.

computron:
Thanks!

PR #201, opened 2016-12-09
warn webgui query about slow queries

dwinston:
If the `FWAPP_PERF_WARNINGS` env var is set, flash a message to the top of the home page after submitting a query that doesn't use an index of the corresponding collection.

Encourages users to e.g. [add to](https://pythonhosted.org/FireWorks/performance_tutorial.html) `user_indices` and `wf_user_indices` in their `my_launchpad.yaml` config file.

### DIFF
--- a/docs/design_tips.rst
+++ b/docs/design_tips.rst
@@ -7,7 +7,7 @@ Tips for designing Firetasks, FireWorks, and Workflows
 Designing Firework specs
 ========================
 
-Recall that the **spec** of a Firework completely bootstraps a job and determines what will run. One the major tasks as a FireWorks user is to decide how your **spec** is structured. We suggest you keep the following suggestions in mind:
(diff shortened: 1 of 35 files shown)


computron:
Thanks!

Can you put the PERF_WARNINGS variable in the fw_config instead of env var?

The only reason we have env vars for username/password in the app is because I couldn't find a better way to communicate this information to the Flask app. But in general, we are trying to move away from environment variables (e.g., how pymatgen removed MAPI_KEY env var, etc).

computron:
@dwinston - see my comments above (just in case using the @ shoutout helps send the message your way)

dwinston:
@computron it's now settable as `perf_warnings: true` line in config YAML.

dwinston:
Okay, now actually using fw_config rather than lpad init (!)

computron:
thanks!

PR #203, opened 2016-12-09
Merge mlaunch into rlaunch command

shyamd:
Added multi command to rlaunch
Ported mlaunch parser to rlaunch


computron:
Thanks! Can you update the docs too? Just search for "mlaunch" within the entire FireWorks repo, and without any file masks so that it includes .rst files, etc.

computron:
thanks!

PR #204, opened 2016-12-09
Qlaunch fw

montoyjh:
* Adds qlaunching of specific fireworks, e. g. `qlaunch -r singleshot -f fw_id`
* Also added a few notes to priority_tutorial.rst about rlaunching and qlaunching specific fireworks.

Note that I changed the abbreviated flag for --fill_mode to -fm to avoid overlap with -f, which I thought should be the same for rlaunch and qlaunch.  They're part of separate parsers, so it's not strictly necessary, but might make things a little less confusing.


computron:
Thanks!

PR #208, opened 2016-12-13
filepad updates

matk86:
* add "compressed" bool key
* delete old file contents on update
* add build_indexes and reset
* add/update unittests

### DIFF
--- a/fireworks/user_objects/firetasks/filepad_tasks.py
+++ b/fireworks/user_objects/firetasks/filepad_tasks.py
@@ -55,7 +55,6 @@ class GetFilesTask(FiretaskBase):
     optional_params = ["filepad_file", "dest_dir", "new_file_names"]
 
     def run_task(self, fw_spec):
-        # TODO: DB should store whether the file was compressed by FilePad, and decompress the file automatically if so
         fpad = get_fpad(self.get("filepad_file", None))
         dest_dir = self.get("dest_dir", os.path.abspath("."))
         new_file_names = self.get("new_file_names", [])
--- a/fireworks/user_objects/firetasks/tests/test_filepad_tasks.py
+++ b/fireworks/user_objects/firetasks/tests/test_filepad_tasks.py
@@ -52,6 +52,9 @@ class FilePadTasksTest(unittest.TestCase):
                          open(os.path.join(dest_dir, new_file_names[0]), "r").read().encode())
         os.remove(os.path.join(dest_dir, new_file_names[0]))
 
+    def tearDown(self):
+        self.fp.reset()
+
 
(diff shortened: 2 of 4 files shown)


computron:
@matk86 Thanks! A couple of notes

1. For the comment "# TODO: set the label=file_id if label == None" I meant to make sure that the root document that was inserted into the database has label == file_id and we are **not** inserting {"label":None} into the database. When you check the database, the label should == file_id if you did not set the label. Please confirm this is the case, i.e., that if you do not set a label, that a non-None value of label corresponding to the file_id is inserted as the label in the filepad document database.

2. For the build_indexes, it seems if I just auto_load the FilePad or use the FilePad_tasks I may never actually build the indexes? Note that if you are concerned about performance (i.e., of creating/ensuring the index every time the FilePad is loaded), I would test whether this actually makes any real difference in time to load the FilePad. Note that I'd much rather take a 20ms hit in performance than end up with a situation where the indexes are not there (which will cause much bigger performance problems down the line, and will also removes the important added db-level check of making sure the label and file_id is not repeated twice leading to a corrupt db)

computron:
(side note: for the LaunchPad itself, a database reset *must* occur before the user inserts anything in the db because otherwise FWS will complain that it cannot get the next fw_id from the counter collection. For FilePad, there is no such guarantee that the collection reset must occur as far as I can tell)

matk86:
1. yes, that is the case. I have a unittest for that too.
2. will update and get back to you

matk86:
added index building and reset if empty..

computron:
For 2. why not just always call build_indexes when instantiating a FilePad rather than counting the number of entries each time and then doing a hard reset depending on the result? 

Note that:
i) The hard reset is kind of a forceful procedure
ii) Creating the index (even when not needed) should be a very lightweight procedure (this used to also be called ensure_index); it will *not* rebuild the index on each call, so it is probably comparable to counting the number of entries in the collection (which you no longer need to do if you just always call build_indexes when creating a FilePad).
iii) If there are multiple threads that are creating FilePad(), e.g., multiple FireWorks running simultaneously, I would be a little less trustworthy of your implementation. There can be a time delay between the count() operation and the delete operations which is not a good thing and in very rare cases can lead to bad behavior. For example, the first process might be deleting the first gridfs collection might take 3 seconds, during which time another process inserts into FilePad (since the length of the documents is now zero, it is all clear to insert), after which the first process deletes the second gridfs collection (including the thing that was just inserted by the the second process)

If you want I can just do this on my end, just wanted to point that out.

matk86:
makes sense, removed count and reset (build_indexes was already there)

computron:
Thanks!

PR #210, opened 2016-12-15
Ensure encoding of UTF-8 before parsing command output

xhqu1981:
It is strange that stdout from SLURM submission command is a raw bytes string. Anyway, I got a raw byte string in a Python3 environment. Now the code checks the type of "output_str" variable, if it is really a raw bytes string, convert it to UTF-8 str explicitly. Otherwise, the split() function will raise an error.


computron:
Can you just skip the "if" statements and always convert? You can put an in-line comment that the decode helps with Py3 compatibility of bytestring

Pretty sure that Py2 code will also work if you decode string to utf-8

xhqu1981:
Done. Thanks.

computron:
Thanks!

PR #212, opened 2017-01-11
Add in MongoDB syntax into update_fws

shyamd:
Fix changelog
Add --mongo flag to use direct mongo syntax in update_fws
Add mongo flag into update_spec


computron:
Thanks!

PR #216, opened 2017-02-23
updates

matk86:
*  plot_wf  bug fixes
*  filepad updates: 
* add count method
* add task to insert files that match the given pattern, add unittest, some cleanup

### DIFF
--- a/fireworks/user_objects/firetasks/filepad_tasks.py
+++ b/fireworks/user_objects/firetasks/filepad_tasks.py
@@ -17,57 +17,41 @@ class AddFilesTask(FiretaskBase):
     A Firetask to add files to the filepad.
 
     Required params:
-        - paths ([str]): list of paths to files to be added
-        - labels ([str]): list of labels, one for each file in the paths list
+        - paths (list/str): either list of paths or a glob pattern string.
 
     Optional params:
+        - identifiers ([str]): list of identifiers, one for each file in the paths list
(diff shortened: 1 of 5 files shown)


computron:
Hi @matk86 

Why not add the pattern matching as an option to the default AddFilesTask rather than having a separate task? Might help remove duplicated portions of code as well as keep the feature set (labels, etc) consistent between the two modes.

computron:
(e.g., a flag called "allow_wildcards" or "pattern_match" in the AddFileTask)

matk86:
i considered that option, since paths and labels are required params  for AddFilesTask, one would be forced to set those even though those are not needed when the paths are to be obtained from the given pattern..

computron:
@matk86 that is easily fixable

1 - the "paths" requirement is a complete non-issue. It is the same exact thing as your patterns input. It is just that when you set the use_wildcards flag to true, then paths becomes interpreted as patterns. It's the same exact parameter
2 - as for labels, just make it an optional parameter in all cases. If it's not set, make sure it is the *full* path of the file to try to avoid collisions since the label is supposed to be unique
3 - (optional) probably, we want rename the word "label" to "identifier". The word label sounds like a category and non-unique, whereas this string is supposed to uniquely identify the file.

With these in place you can just have a single FireTask

computron:
Thanks!

PR #217, opened 2017-02-28
Changed checks in Workflow.state to use generator expressions

richardjgowers:


fireworks/core/firework.py
```diff
@@ -735,21 +735,21 @@ def state(self):
         m_state = 'READY'
         #states = [fw.state for fw in self.fws]
         states = self.fw_states.values()
-        leaf_states = [self.fw_states[fw_id] for fw_id in self.leaf_fw_ids]
-        if all([s == 'COMPLETED' for s in leaf_states]):
+        leaf_states = (self.fw_states[fw_id] for fw_id in self.leaf_fw_ids)
+        if all(s == 'COMPLETED' for s in leaf_states):
```
richardjgowers: So without the square brackets `all` can exit as soon as the first False value occurs, rather than creating the entire list and then checking items one by one

fireworks/core/firework.py
```diff
@@ -735,21 +735,21 @@ def state(self):
...
+        elif any(s == 'PAUSED' for s in states):
             m_state = 'PAUSED'
-        elif any([s == 'FIZZLED' for s in states]):
+        elif any(s == 'FIZZLED' for s in states):
             # When _allow_fizzled_parents is set for some fireworks, the workflow is running if a
             # given fizzled firework has all its childs COMPLETED, RUNNING, RESERVED or READY.
             # For each fizzled fw, we thus have to check the states of their children
-            fizzled_ids = [fw_id for fw_id, state in self.fw_states.items()
-                           if state not in ['READY', 'RUNNING', 'COMPLETED', 'RESERVED']]
+            fizzled_ids = (fw_id for fw_id, state in self.fw_states.items()
```
richardjgowers: Using generator expressions means if the first `fw_id` caused the break below, only one would have to be retrieved and checked

computron:
Thanks!

PR #218, opened 2017-03-07
Added an option to redirect stdout and stderr to the launch directory

shyamd:
Redirect stdout to fws.out and stderr to fws.error in the launch directory
Available for rlaunch rapidfire and rlaunch multi

### DIFF
--- a/fireworks/core/rocket.py
+++ b/fireworks/core/rocket.py
@@ -22,6 +22,7 @@ from fireworks.fw_config import FWData, PING_TIME_SECS, REMOVE_USELESS_DIRS, PRI
     PRINT_FW_YAML, STORE_PACKING_INFO
 from fireworks.utilities.dict_mods import apply_mod
 from fireworks.core.launchpad import LockedWorkflowError
+from fireworks.utilities.fw_utilities import get_fw_logger
 
 __author__ = 'Anubhav Jain'
 __copyright__ = 'Copyright 2013, The Materials Project'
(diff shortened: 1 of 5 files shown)


shyamd:
Found some bugs. going to fix and then reopen.

computron:
@shyamd two notes

1. Is this ready to merge now?
2. The default out/error name used in the queue adapter templates in FW_job.out and FW_job.error. Although this is a very soft standard, it might be worth using that name instead of "fws.error" to stay consistent. What do you think?

computron:
@shyamd Also see the Travis unittest failures:

AttributeError: 'NoneType' object has no attribute 'get_logdir'

xhqu1981:
I am just wondering why it can be so complicated. If I understand correctly, the purpose of this pull request is to dispatch the output to individual files for each process. Simply config the logging file name to process dependent would do the work, right? e.g. append process id to line 50 in fw_utilities.py.

As for stderr, if it is from VASP, it is better to redirect in the workflow. If it is from specific python code in FireWorks, it should be possible to forward the context in that place.

I didn't look into it deeply, I might wrong. Please let me know if config logging file name doesn't work.

shyamd:
@xhqu1981 
It's actually not too complicated. This is what is involved

1.) Redirect stdout and stderr, which is what the context manager does.
2.) Add in options to make this change
3.) Change the logging structure in rocket so that it uses rocket.launcher rather then the launchpad logger

1.) and 2.) are pretty obvious. 3.) is needed to ensure that logging in sub-proceses goes to the stdout from that subprocess rather than the stdout of the root process.

xhqu1981:
OK. I am fine with the current solution.

shyamd:
@computron 
Fixed the file names and the travis unittest error
This is ready to merge now

computron:
@shyamd the file names look the same to me? i.e. fws.out instead of FW_job.out

shyamd:
Forgot to push those updates properly. Now it should look right.

computron:
Thanks!

PR #219, opened 2017-03-23
Fix bug in webgui query filtering

dwinston:
Avoids KeyError if one of two session vars for firework/workflow query filtering is not set.


computron:
Thanks!

PR #220, opened 2017-04-11
Added qlauncher option for remote shell.

dotsdl:
For at least one cluster I've encountered (PSC Bridges), the [default shell used by fabric](http://docs.fabfile.org/en/1.13/usage/env.html#shell) was not initializing with the configuration files I expected to give the proper environment for my qlauncher. This PR adds an option to the qlauncher for passing in the shell command that fabric will use.


computron:
Thanks!

PR #221, opened 2017-04-11
``--max_waiting`` for qlauncher

dotsdl:
In addition to the ``--maxjobs_queue`` option for ``qlaunch`` which limits the total number of jobs in the queue for the given user, it's been useful to have a ``--maxjobs_waiting`` parameter limiting the number of jobs further on the basis of how many are in a waiting state. 

For cases in which a cluster has multiple partitions/queues, but a global limit on the number of jobs that can be waiting/running at any given time, this option can be used to ensure that one queue isn't filling up this quota with jobs that have to wait a long time.

The underlying parsing of queue status output is implemented for SLURM and SGE. We don't have an implementation of this for other resource managers, though, yet.


computron:
Hi @dotsdl 

Thanks for implementing! I think this is a useful feature but I would suggest a few minor code restructuring changes that might help reduce the amount of new code and new methods that are needed for its implementation.

Currently, you add a new method to queue adapters, get_njobs_waiting. Is it possible to get remove of this method and simply add the "waiting" constraint as another parameter to the existing
get_njobs_in_queue()? The latter already has a constraint for username, and it seems logical to keep other constraints for the number of jobs in the queue here rather than create new methods for different types of constraints.

Note that if you don't have an implementation of the waiting code for certain queue systems but not others, just return an error (e.g. ValueError) *if* the waiting constraint is set but you don't have an implementation. For those queue systems, people will have to run without the waiting constraint (like they already do) or will quickly find out they need to implement the waiting constraint if they want it for that queue system.

Similarly, the internal method _get_number_of_jobs_waiting looks at first glance to be able to implemented parameter to the existing _get_number_of_jobs_in_queue method rather than a new method.

What do you think of these proposed changes? Are they workable?

computron:
@dotsdl any thoughts?

computron:
Since the last active comment was April 14, I am closing this PR. Please feel free to re-open if this has new updates.

PR #222, opened 2017-04-12
add gres support to  SLURM_template.txt

specter119:
add Specifies a comma delimited list of generic consumable resources to SLURM_template.txt, to support manage the occupation of GPU.


computron:
Thanks!

PR #223, opened 2017-04-15
Rerun single firework

shyamd:
Added the ability to rerun a single firework and not its children

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -841,13 +841,14 @@ class Workflow(FWSerializable):
 
         return list(set(updated_ids))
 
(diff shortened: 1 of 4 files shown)


computron:
I am afraid this is going to do more harm than good. The workflow is a set of dependencies, and it doesn't make sense that (say) a child is COMPLETED but the parent is READY/RUNNING if the child is a dependency of the parent. For example, if I were to accept this commit, then someone down the line might try to implement a speedup that says a Workflow is COMPLETED if all the leaf children are COMPLETED, which should be a valid algorithm, but this would break that.

The better way to "rerun" a single Firework would be to have a method that copies that single Firework into a new independent Workflow.

PR #224, opened 2017-04-21
fix: always define variables used in except block

gpetretto:
If an error is raised before ping_stop and btask_stops have been defined the except block fails due to the variable being referenced before assignment. This PR fixes the problem78.


computron:
Thanks!

PR #225, opened 2017-06-05
add remove_fws method to Workflow

matk86:
* remove fireworks corresponding to the given fw ids
* added tests
* note: was orginally in atomate


computron:
Thanks!

Note: I usually try to avoid deepcopy

PR #226, opened 2017-07-18
WIP: Add protocol for file transfer.

shyuep:
Do not merge yet. No tests are written, and certain functionality not done yet. But this is a draft implementation of basic functionality for comment. 

A simple example script to demo is below:

```
from fireworks import ScriptTask
from fireworks.core.firework import Firework, Workflow
from fireworks.core.launchpad import LaunchPad
from fireworks.core.rocket_launcher import rapidfire

from fw_tutorials.dynamic_wf.printjob_task import PrintJobTask

# set up the LaunchPad and reset it
launchpad = LaunchPad()
launchpad.reset('', require_password=False)

# create the Workflow that passes job info
fw1 = Firework([ScriptTask.from_str('echo "This is the first FireWork" > test')], spec={"_pass_job_info": True, "_output_files": {"fwtest": "test"}}, fw_id=1)
fw2 = Firework([ScriptTask.from_str('cat fwtest')], fw_id=2, 
    spec={"_input_files": ["fwtest"]})
wf = Workflow([fw1, fw2], {fw1: [fw2]})

# store workflow and launch it locally
launchpad.add_wf(wf)
rapidfire(launchpad)
```

Comment on whether the general implementation is fine first. Then I will do the other magical things like auto detection of zip formats and add some basic tests in subsequent commits.

@montoyjh @matk86 @computron especially to comment.


shyuep:
PS This took all of 15 min to write and test.

shyuep:
ONe thing I am truly confused about. it seems that Firework has a "parents" attribute. But no matter what value I pass in there, the attribute is always empty. As far as I can tell, the to_dict and from_dict completely ignores this attribute, which means parent information is lost. Am I using it wrong?

shyuep:
Changed the spec a little. Now zipped files are supported transparently. New example:

```
from fireworks import ScriptTask
from fireworks.core.firework import Firework, Workflow
from fireworks.core.launchpad import LaunchPad
from fireworks.core.rocket_launcher import rapidfire

from fw_tutorials.dynamic_wf.printjob_task import PrintJobTask

# set up the LaunchPad and reset it
launchpad = LaunchPad()
launchpad.reset('', require_password=False)

# create the Workflow that passes job info
fw1 = Firework([ScriptTask.from_str('echo "This is the first FireWork" > test1')], spec={"_pass_job_info": True, "_files_out": {"fwtest1": "test1"}}, fw_id=1)
fw2 = Firework([ScriptTask.from_str('gzip fwtest1')], fw_id=2, parents=[fw1],
    spec={"_files_in": ["fwtest1"], "_files_out": {"fwtest2": "fwtest1.gz"}})
fw3 = Firework([ScriptTask.from_str('cat fwtest2')], fw_id=3, parents=[fw2],
    spec={"_files_in": ["fwtest2"]})
wf = Workflow([fw1, fw2, fw3], {fw1: [fw2], fw2: [fw3]})

# store workflow and launch it locally
launchpad.add_wf(wf)
rapidfire(launchpad)
```

computron:
Note: The ```parents``` attribute is used when the Firework is part of a workflow - it gets used to create the workflow links. i.e., the parents of a Firework is really a workflow property. You'll see it used if you put the Firework inside a Workflow.

Overall I think the syntax is on a good track. Some notes:

* As I understand it, "_output_files" is the key used to determine what files should be pushed on to the next FW. But then the rocket pushes that *same* key to the child FW! This forces the child FW to also pass the same _output_files to its own children (from my reading of the code). The parent should not be forcing the _output_files from the child? i.e.  ```{"_push_all": {"_output_files": filepaths}})``` should read ```{"_push_all": {"_output_files_parents": filepaths}})```
* Atomate has a FileClient that can be used to do remote file transfers as well as local ones. Folding this into FireWorks would be nice
* Note that your current implementation will always pull the files in the beginning of a rocket.run(). However, you usually do not want to do this if the "_recover_launch" option is set for task_level recovery. In that case, you are starting from (say) the third task down and pulling the files again could lead to bad behavior (e.g., overwriting files that got generated in the first or second task)
* It would be nice if one could use glob to search for a file path rather than explicit name. One could set either the first or last sorted hit to match
* I would make sure this framework behaves well on rerun. e.g., run a completed Workflow, rerun it from the beginning, and run it again. Then check the ``_output_files``, etc. keys to make sure they make sense and are not duplicated.

Thanks!

shyuep:
Some quick responses:

> As I understand it, "_output_files" is the key used to determine what files should be pushed on to the next FW. But then the rocket pushes that same key to the child FW! This forces the child FW to also pass the same _output_files to its own children (from my reading of the code). The parent should not be forcing the _output_files from the child? i.e. {"_push_all": {"_output_files": filepaths}}) should read {"_push_all": {"_output_files_parents": filepaths}})

I fixed this in the new syntax. Now the files are actually pushed to prev_files. I also changed the syntax a bit.

> Atomate has a FileClient that can be used to do remote file transfers as well as local ones. Folding this into FireWorks would be nice

Ok, this I will leave to someone else to move this to Fireworks.

> Note that your current implementation will always pull the files in the beginning of a rocket.run(). However, you usually do not want to do this if the "_recover_launch" option is set for task_level recovery. In that case, you are starting from (say) the third task down and pulling the files again could lead to bad behavior (e.g., overwriting files that got generated in the first or second task)

Fixed in latest push.

> It would be nice if one could use glob to search for a file path rather than explicit name. One could set either the first or last sorted hit to match

Done.

> I would make sure this framework behaves well on rerun. e.g., run a completed Workflow, rerun it from the beginning, and run it again. Then check the _output_files, etc. keys to make sure they make sense and are not duplicated.

Tested and seems to work fine. Let me know if this is not the case.

P.S. there are some failing tests, but that is mainly because prev_files is being set. But as far as I can tell, there is no substantive issues.

shyuep:
Tests added. Should be mergable unless you have further input.

shyuep:
I also added docs. THis is all I am contributing for now. Go ahead and edit and merge as necessary. Future coders feel free to add on functionality.

computron:
Thanks! Also for including some docs in the tutorial

PR #227, opened 2017-08-01
Task-level checkpointing

montoyjh:
This PR refactors task-level checkpointing to occur in normal operation of the firework, rather than just the recovery.  Task-level checkpointing of the accumulated mod_spec, update_spec, stored data and task number occurs at the beginning of the execution of each firetask in the rocket.run method and is stored in the state_history.  This enables the following:

* If a firework is stopped on the queue due to job cancellation or encountering the walltime limit (or is lost for any other reason), it can still be recovered from the state_history, and will be on `lpad rerun_fws --task-level`.
* Task-level reruns in the case of exceptions is still supported, the recovery information is just not stored in the action key (stored in the state history of the launch).
* If a firework is not explicitly rerun in task-level mode, the recovery information is removed by default from the firework.
* LaunchPad.rerun_fws_task_level's functionality is merged into LaunchPad.rerun_fw.

## TODO:
* In the course of doing this, I fell down a rabbit hole trying to refactor rocket.py.  I've maintained the branch in which I'm doing this, but wanted this specific checkpointing functionality sooner for my work.  Basically, that branch aimed to reduce a lot of code duplication related to the offline/online functionality split by carrying the launch through the rocket.run method and dumping it to file (rather than having separate files for FW_offline/FW_ping etc.).  I ran into a lot of issues related to interdependency and threading that I wasn't sure how to fix, but might have more time to tackle later.


computron:
thanks!

PR #231, opened 2017-08-28
Enable optional queue setting in SGE queue adapter

utf:
Currently the SGE queue adapter requires the queue parameter to be set. This causes an issue for clusters where SGE is configured such that jobs are assigned a queue at run time, as it means the number of queuing jobs cannot be determined.

For these systems it is OK if the queue parameter is not set, therefore I have made this an optional parameter (similar to how it is handled in the SLURM section of the code).


computron:
ok thanks! note that I might shorten the code to if self.get("queue")

PR #234, opened 2017-09-05
rapidfire until completion

### LINKED ISSUE #233: Sleeping rapidfire works in diamond workflows.
So I'm running workflows which occasionally have to join then fork back outwards, so as a trivial example..
```
FW1
V
FW2 FW3
V
FW4
V
FW5 FW6
```
If I run two parallel processes of `rlaunch rapidfire`, then one will do `FW1`, and the other will exit as it sees no job immediately able to run.  The problem being I want the other Worker to stay alive so that `FW2` and `FW3` can be done in parallel.

If I instead do two parallel processes of `rlaunch rapidfire --nlaunches infinite` then both workers will stay alive and perform the workflow as I want, but they will live indefinitely, as once all FWs are done, they will get stuck in a loop of sleeping.

I think what is needed is for `rlaunch rapidfire` to behave more like `rlaunch multi`, in which parallel processes will sleep (and stay alive) if they see that there are future jobs that will appear.  The check which enables this is here:

https://github.com/materialsproject/fireworks/blob/master/fireworks/features/multi_launcher.py#L87

So with `nlaunches=0` (ie until 'completion'), a `rapidfire` Worker should stay alive until it sees that there are no more **future jobs possible** for it to complete.

richardjgowers:
Fixes #233 

Makes rapidfire with nlaunches=0 only quit once there are no more READY or WAITING launches

### DIFF
--- a/fireworks/core/launchpad.py
+++ b/fireworks/core/launchpad.py
@@ -583,6 +583,31 @@ class LaunchPad(FWSerializable):
         q = fworker.query if fworker else {}
         return bool(self._get_a_fw_to_run(query=q, checkout=False))
 
+    def future_run_exists(self, fworker=None):
+        """Check if database has any current OR future Fireworks available
+
+        Returns:
+            bool: True if database has any ready or waiting Fireworks.
+        """
+        if self.run_exists(fworker):
+            # check first to see if any are READY
+            return True
+        else:
+            # grab all [RUNNING/RESERVED] fireworks
+            q = fworker.query if fworker else {}
+            q.update({'state': {'$in': ['RUNNING', 'RESERVED']}})
+            active = self.get_fw_ids(q)
+            # then check if they have WAITING children
+            for fw_id in active:
+                children = self.get_wf_by_fw_id_lzyfw(fw_id).links[fw_id]
+                if any(self.get_fw_dict_by_id(i)['state'] == 'WAITING'
+                       for i in children):
+                    return True
+            else:
+                # if we loop over all active and none have WAITING children
+                # there is no future work to do
+                return False
+
     def tuneup(self, bkground=True):
         """
         Database tuneup: build indexes
--- a/fireworks/core/rocket_launcher.py
+++ b/fireworks/core/rocket_launcher.py
@@ -85,10 +85,14 @@ def rapidfire(launchpad, fworker=None, m_dir=None, nlaunches=0, max_loops=-1, sl
     start_time = datetime.now()
     num_loops = 0
 
-    while num_loops != max_loops and (not timeout or (datetime.now() - start_time).total_seconds() < timeout):
+    def time_ok():
+        # has the rapidfire run timed out?
+        return (timeout is None or
+                (datetime.now() - start_time).total_seconds() < timeout)
+
+    while num_loops != max_loops and time_ok():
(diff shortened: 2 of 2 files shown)


computron:
Thanks, but there is still an issue here in that a Firework can be WAITING even if its parent is FIZZLED or DEFUSED. In that case, it may be WAITING forever, i.e., never become READY, and your rapidfire process will continue forever.

Ideally we'd have a different state for Fireworks that are WAITING behind a READY/RUNNING Firework (still currently expected to run soon) and Fireworks that are WAITING behind a FIZZLED/DEFUSED Firework (may have "gone cold" and never be ready to run unless the user fixes the parent or reignites the parent in the future) but it's not the case at the moment.

If future_run_exists() took that distinction into account it could prevent that "forever waiting for a WAITING firework" issue.

richardjgowers:
@computron ok thanks, I'll see if I can fix it so it will only finds possible WAITING fireworks.  I'm guessing it'll be a little bit too much to try and add a new state just for this unless this is something which would be useful for other things?

fireworks/core/launchpad.py
```diff
@@ -583,6 +583,33 @@ def run_exists(self, fworker=None):
...
+            # check first to see if any are READY
+            return True
+        else:
+            # grab all [RUNNING/RESERVED] fireworks
+            q = fworker.query if fworker else {}
+            q.update({'state': {'$in': ['RUNNING', 'RESERVED']}})
+            active = self.get_fw_ids(q)
+            # then check if they have WAITING children
+            for fw_id in active:
+                children = self.get_fw_by_id(fw_id).links[fw_id]
```
richardjgowers: I'm not sure if this is the best/lightest way to do this query.  I want to check the state of all children of fireworks with state of RUNNING or RESERVED

richardjgowers:
I've reworked how the future check works.  Instead of going through all WAITING FWs, it probably is quicker (assuming that you have more WAITING than RUNNING, which is probably true for most reasonable WFs) to check to see if behind any RUNNING there is a waiting FW.

fireworks/core/launchpad.py
richardjgowers: I.e. can I access `.links` without having to deserialize the Firework

computron:
Thanks! A few minor notes

* I didn't see the reason to add a ``state`` parameter in get_a_fw_to_run(). I think that should be removed (even your proposed docs suggest not to use any other value than READY).

* For time_ok() it can just be one-liner:
``
return timeout is None or (datetime.now() - start_time).total_seconds() < timeout
``

* Similarly I think it's clear enough to just write: ``if nlaunches == 0 and not launchpad.future_run_exists():`` rather than the nested if

Let me know what you think

computron:
Thanks! I will merge and perhaps make some final cleanups, let me know if you see any issues afterwards

richardjgowers:
@computron ok I've tidied things up.  Your last point wouldn't work cleanly, because it will enter the `elif num_launched == nlaunches:` clause and exit prematurely, you'd have to do something like `if not nlaunches == 0 and (nlaunches == num_launches):`

computron:
Ok thanks, I had made the change but see why it's incorrect and reverted my modification. The logic is getting tricky ...

Hopefully the most recent version works for you

PR #237, opened 2017-10-11
Extended TLS/SSL configuration in launchpad file

ikondov:
The existing SSL option in configuration file allows encrypted database connection, however it is insufficient if the server requires client authentication with a client certificate. This is the reason for this extension proposal. The change has been tested with local MongoDB servers in the group with and without SSL. No changes in the launchpad file are necessary for the non-SSL mode, for the previously implemented SSL mode (CA cert ) the ``ssl`` has to be set to ``true``. **NOTE**: The key ``ssl_ca_file`` was changed to ``ssl_ca_certs`` for more consistency with pymongo keys but the key name can be reverted before merging if there are many configurations in production which can get broken. However, the previous SSL support was not documented so that this might not be necessary.


computron:
Thanks! Regarding the renaming, I am just going to make the from_dict() of LaunchPad backward-compatible shortly after merging. That way, the old LaunchPad files will still work but new LaunchPad files that are created will follow the new format.

PR #239, opened 2017-10-12
Firetasks for dataflow management

ikondov:
This is a bunch of firetasks providing methods (elaborating on ``FWAction`` and the basic ``PyTask``) to stage data and metadata from fireworks to their children. The most helpful ones from this set are
 * ``PythonFunctionTask`` for integrating a python function which is **unaware** of FireWorks and does not have to return a ``FWAction`` object.
 * ``CommandLineTask`` to integrate a command line with full set of command line flags and options for managing input/output from/to files and to/from stdin/stdout. Again, the command is unaware of  FireWorks and does not have to return a ``FWAction`` object.
* ``ForeachTask`` splits an array into chunks and inserts fireworks to process each chunk. This is very useful for distributing high throughput (single program - multiple data) jobs over resources.

These and the other proposed Firetasks are described in the documentation under *Using the Dataflow Firetasks*. 

This pull request is a work in progress. Tests will be added very soon.


computron:
@ikondov Thanks!

My only hesitation in merging this is to try to avoid confusion / problems for users on things like distinction between PyTask and PythonFunctionTask or ScriptTask and CommandLineTask. There should ideally be just one of each and that one would encompass all use cases.

For PyTask and PythonFunctionTask, it seems there should be a way to modify PyTask to do what you want without too much effort. Perhaps just allowing inputs to be an alternate for args and having an option to return things in a certain format? I think it will be a much better situation to have one task rather than two different ones. Note that PyTask doesn't require you to return a FWAction, and there is already a "stored_data_varname" that helps you return a custom FWAction. You could another custom parameter that helps PyTask return the type of FWAction you want for data processing.

For ScriptTask vs CommandLineTask, this one looks a bit harder to merge one into the other. But, if you had thoughts on how to do that, it would be better to have just one.

What do you think?

ikondov:
@computron Thanks a lot for comments! I agree, it is in fact feasible to move the implementation of ``PythonFunctionTask`` into ``PyTask`` and extend the `PyTask` interface with three optional parameters: inputs, outputs and chunk_number. I will do this in the next commit. 

Regarding ``CommandLineTask``: also here (with some more effort) it is possible to insert the new code into ``ScriptTask`` but we should be aware of that the ``CommandLineTask`` starts a command available on the system with user-defined arguments (in the dict ``command_spec``) while ``ScriptTask`` starts a user-defined script, two quite different things. After all, the ``ScriptTask`` will become a bit more complex (~200 lines longer). I can redesign and test in another branch and also think of a new task name better matching both use cases. What do you think?

computron:
Ok great! I would vote to merge PyTask for sure. ScriptTask I will leave up to you. However, you should maintain backward compability in all cases as these are commonly used tasks.

ikondov:
As suggested I extended the PyTask without changing behavior with static arguments (full backward compatibility). Wrote unit tests for all new Firetasks and the new use cases of PyTask, and extended the documentation correspondingly. I think now it is ready for merge.

computron:
Thanks! I will merge now. I might make some minor cleanups here and there after merging.

PR #240, opened 2017-10-16
Performance for LaunchPad.add_wf

richardjgowers:
I was getting annoyed with submitting a ~50,000 firework Workflow was, so I made a few changes which seem to work for me, but maybe there's something obvious I missed.

 - Added optional quantity kwarg to `LaunchPad.get_new_fw_id` (allows multiple fw_ids to be taken at once)
 - `LaunchPad._upsert_fws` with `reassign_all = True` now does fewer database transactions (3 in total, rather than N)


computron:
Awesome thanks!

Our workflows are always <50 Fireworks so we are almost always unaware of performance problems at the 50,000 scale ...

richardjgowers:
@computron no problem, thanks for reviewing

PR #241, opened 2017-10-19
Rewrote Workflow.Links to maintain parent_links rather than build on demand

richardjgowers:
This improves performance a little bit, as building the `parent_links` dict is slower with large Workflows.

The API of the Links object should be identical to before.

There was a weird bug in `Workflow.__init__` where when creating the WF from the database, the keys in `links_dict` were all strings, so the check "`if fw.fw_id not in links_dict and fw not in links_dict:`" would always think that the FW wasn't in the links_dict (because it was a string) so you'd end up with a empty list for all FWs, (as well as the key-as-string entry), which would later get ironed out in `Links.__init__`.  This is why the usage of Links is slightly changed there.

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -708,6 +708,7 @@ class Workflow(FWSerializable):
         name = name or 'unnamed WF'  # prevent None names
 
         links_dict = links_dict if links_dict else {}
+        self.links = Workflow.Links(links_dict)
 
         # main dict containing mapping of an id to a Firework object
         self.id_fw = {}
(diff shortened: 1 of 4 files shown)


computron:
Note - the tests are failing

The string thing might be because MongoDB (unlike Python) does not allow keys of dictionary to be ints, so they need to be string

richardjgowers:
@computron yeah I think I've missed something, I'll look into it

computron:
I am closing this PR for the moment since it doesn't seem active. Please re-open if you're able to resolve the issues with the tests.

Issue #242, opened 2017-10-27
test_command_line_task_3 doesn't work on Mac


computron:
@ikondov 

test_command_line_task_3 appears to use a "date" command that only works on Unix systems. It fails on Mac with error:

RuntimeError: b'date: illegal option -- -\nusage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... \n            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]\n'

For now I have commented out the unit test. Can you decide the best thing to do? e.g., remove unit test altogether, modify unit test to not use commands that work on Mac, or perform some kind of test of your environment and skip the unit test if you're on a different system?

ikondov:
Thanks a lot for reviewing and merging!

In this case I would suggest to omit this test. The problem is due to the incompatibility between GNU core utilities and the Darwin shell utilities. They have the same names (date, stat, touch) but different command line options and/or output formats. Unfortunately i cannot adapt it to work because I have no Mac.

I am very sorry for this. I did not know that the software should work on Mac OS. Probably we have to extend the docs and contribution guidelines with this detail and add some platform-specific CI tests?

computron:
No worries! I merged the change and added a new statement about OS compatibility to the contrib guidelines.

To give a little background, some people use FireWorks to automate tasks on their laptop (including Windows systems). We are trying to also support that use case along with the supercomputing use case.

PR #243, opened 2017-10-29
skip test_command_line_task_3 when platform is not GNU/Linux

ikondov:
This is a solution to Issue https://github.com/materialsproject/fireworks/issues/242. I simply skip the test on a non GNU/Linux platforms.


computron:
Thanks! I am just going to change the ``return`` to raise a ``SkipTest`` instead after merging

PR #244, opened 2017-11-07
Recovery fix

montoyjh:
Task-level recovery has a bug in that LaunchPad.rerun_fws unsets the wrong spec key on a normal rerun.  This PR should fix.  Might want to wait on tests, something was a bit weird on my circle tests, but I think it's a config issue on my fork.


computron:
Thanks!

PR #245, opened 2017-11-08
correction for errors when handling multiple qadapters on a single LS…

jboes:
…F cluster

When running multiple queue adapters on a single LSF cluster, the maximum queue count for each rapidfire run used to refer to the total number of submitted jobs. This is a potential problem for automation since not all queues completed jobs at the same rate or are equally available. After a while, the queues which complete jobs faster will deplete their supply of queued jobs which will steadily be replaced by queue submissions to the slower queues. Eventually, submissions to the slowest queue will beat the others out.

This is a fix for LSF systems which identifies the number of jobs submitted to a specific queue. That way, submission can be managed across all queues correctly.


computron:
Thanks!

PR #247, opened 2017-11-21
Remove hard-coded secret key

mkhorton:
I don't think it matters too much right now, but we should probably avoid hard-coding a secret key in a public repo? I think this is mostly used for session cookies and the like.


computron:
Thanks!

PR #248, opened 2017-11-27
new sub-commands to lpad

ikondov:
Very often we need to extend existing workflows and for that we initially used the python interface to the method `fireworks.core.firework.Workflow.append_wf` via a three-line separate python script. Much more convenient would be to use the `lpad` command for this rather general operation. In addition, often a workflow has to be dumped from the LaunchPad for editing or sharing. For these operations, two new sub-commands in the `lpad` command are suggested: `append_wflow` and `dump_wflow`. The added `lpad` help strings document the usage of the new commands and also the tests in `cmd_line_test.sh` are extended accordingly.

### DIFF
--- a/fireworks/scripts/lpad_run.py
+++ b/fireworks/scripts/lpad_run.py
@@ -165,6 +165,21 @@ def add_wf(args):
         lp.add_wf(fwf)
 
 
+def append_wf(args):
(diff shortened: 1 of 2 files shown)


computron:
Thanks!

PR #249, opened 2017-11-28
flask_site: Remove unnecessary, performance-impacting script fetch

dwinston:
The fetched script (livereload.js) depends on a running LiveReload server, which is used only for development and can impact deployments (a server may not quickly 404 the file in part because the request is to a non-standard port, so the page may take tens of seconds to load).


computron:
Thanks!

Issue #251, opened 2017-11-29
Question regarding default SLURM template


mkhorton:
Wondering why `partition` is mapped to `queue` here:

https://github.com/materialsproject/fireworks/blob/92cc1a6f1bfc1faa3b93cdfb7b31b152032156e3/fireworks/user_objects/queue_adapters/SLURM_template.txt#L10

As a matter of book-keeping, it seems like Fireworks template variables should have the same name as the corresponding SBATCH variable (which they do for all the other variables, it's just `partition` that's different).

I'm wary to 'fix' this since I know it would break existing qadapters that are in use by people...

computron:
There are two ways to try set this up:

(i) match SBATCH variables with FireWorks template names
(ii) try to use a consistent FireWorks template name for the same "entity" in different qadapters - e.g., ``-q`` in PBS and ``--partition`` in SLURM are the same entity and both use the FWS name ``queue``.

Unfortunately, I thinks sometimes the template variable names lean towards (i), sometimes they lean towards (ii), and sometimes they just seem to be semi-random.

It would be possible to maintain backwards compatibility and do what you want by having a duplicate line for partition like this:
```
#SBATCH --partition=$${queue} 
#SBATCH --partition=$${partition} 
```

Then you could use either variable. But this will probably just makes things worse rather than better.

mkhorton:
Hmm, that makes sense. Perhaps instead we could have a user warning if a variable is specified in the qadapter isn't actually used in the template file?

PR #252, opened 2017-11-29
Add user warning to qadapter for missing keys

mkhorton:
Addresses #251


computron:
Thanks! Can you just have the warning message also print the location of the template file? Otherwise the user needs to take another step to track down the template file ...

mkhorton:
Good point! Added.

PR #254, opened 2017-12-06
Changes to lpad argparser to allow comma separated lists when using -…

sivonxay:
…i and --fw_id

### DIFF
--- a/fireworks/scripts/lpad_run.py
+++ b/fireworks/scripts/lpad_run.py
@@ -651,7 +651,7 @@ def lpad():
     # This makes common argument options easier to maintain. E.g., what if
     # there is a new state or disp option?
     fw_id_args = ["-i", "--fw_id"]
-    fw_id_kwargs = {"type": int, "nargs": "+", "help": "fw_id"}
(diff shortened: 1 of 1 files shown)


computron:
@sivonxay Thanks! But this seems to break the unit tests. Can you see what's going on?

fireworks/scripts/lpad_run.py
```diff
@@ -702,7 +702,7 @@ def lpad():
     addwf_parser.set_defaults(func=add_wf)
 
     append_wf_parser = subparsers.add_parser('append_wflow', help='append a workflow from file to a workflow on launchpad')
-    append_wf_parser.add_argument('-i', '--fw_id', type=int, nargs='+', help='parent firework ids')
+    append_wf_parser.add_argument(*fw_id_args, type=fw_id_kwargs["type"], nargs=fw_id_kwargs["nargs"], help='parent firework ids')
```
computron: I think this code will be easier to read if you just hard-code the ``type`` and ``nargs`` rather than refer to the dict keys

computron: (nevermind, this is fine)

PR #256, opened 2018-01-11
Validate and visualize workflows

ikondov:
I propose a tool to validate workflows in FireWorks, in particular to check that the workflow graph is acyclic and fully connected, the links are valid, and the data flow (if any) is correctly defined. In addition, the tool makes nice workflow graphs using DOT which can be further converted to PDF for viewing with the 'dot' tool. The new class derives from the ``Graph`` class from the igraph package. The workflows can be checked when added to launchpad with the ``lpad add -c`` or ``lpad add --check`` commands and workflows on launchpad can be checked and graphed with the ``lpad check_wflow`` command. The unit tests cover the use cases.


computron:
Thanks!

computron:
Note that I will move python-igraph to an optional dependency, since it could not simply "pip" install on Mac and I'd like to keep the base FireWorks installation as foolproof as possible on all OS

ikondov:
This is okay, I also had a doubt but the utility is really optional. Thanks a lot!

PR #258, opened 2018-01-16
Add optional `lpad webgui` argument for number of server processes

dwinston:
This is important for running multiple `lpad webgui --server_mode` instances on a single host. It's also just a good idea to allow this argument in addition to the `host` and `port` params for the underlying `gunicorn` call.


computron:
THanks!

PR #260, opened 2018-02-08
Bulk insert add wfs

montoyjh:
I've found myself in need of a more efficient method of workflow addition to the launchpad lately (e. g. for adding thousands of workflows at a time), so I wrote a method that does it a bit more efficiently using some of mongo's bulk insertion functionality.

I'm not really sure if this is a good idea for the master branch (it seems less secure than adding them one at a time), but I figured I'd submit this PR to see if there was any feedback on how it might be adapted for general use (or if there's a better strategy), if there are any issues.

Also included rudimentary unit test.

### DIFF
--- a/fireworks/core/launchpad.py
+++ b/fireworks/core/launchpad.py
@@ -315,7 +315,7 @@ class LaunchPad(FWSerializable):
         self.m_logger.info('Added a workflow. id_map: {}'.format(old_new))
         return old_new
 
-    def add_wfs(self, wfs):
+    def bulk_add_wfs(self, wfs):
         """
         Adds a list of workflows to the fireworks database
         using insert_many for both the fws and wfs, is
@@ -351,9 +351,9 @@ class LaunchPad(FWSerializable):
 
         # Insert all fws and wfs, do workflows first so fws don't
         # get checked out prematurely
-        self.workflows.insert_many([wf.to_db_dict() for wf in wfs])
-        all_fws = chain.from_iterable([wf.fws for wf in wfs])
-        self.fireworks.insert_many([fw.to_db_dict() for fw in all_fws])
+        self.workflows.insert_many(wf.to_db_dict() for wf in wfs)
+        all_fws = chain.from_iterable(wf.fws for wf in wfs)
+        self.fireworks.insert_many(fw.to_db_dict() for fw in all_fws)
         return None
 
     def append_wf(self, new_wf, fw_ids, detour=False, pull_spec_mods=True):
@@ -1370,7 +1370,7 @@ class LaunchPad(FWSerializable):
             set_spec = {'$set': {'spec._recovery': recovery}}
             if recover_mode == 'prev_dir':
                 prev_dir = self.get_launch_by_id(recovery.get('_launch_id')).launch_dir
-                set_spec['$set']['spec._launch_dir'] = prev_dir 
+                set_spec['$set']['spec._launch_dir'] = prev_dir
             self.fireworks.find_one_and_update({"fw_id": fw_id}, set_spec)
 
         # If no launch recovery specified, unset the firework recovery spec
@@ -1583,7 +1583,7 @@ class LaunchPad(FWSerializable):
                                                             })
                     if f:
                         self._refresh_wf(fw_id)
-                
+
                 if 'checkpoint' in offline_data:
                     m_launch.touch_history(checkpoint=offline_data['checkpoint'])
                     self.launches.find_one_and_replace({'launch_id': m_launch.launch_id},
--- a/fireworks/core/launchpad.py
+++ b/fireworks/core/launchpad.py
@@ -13,6 +13,8 @@ import random
 import time
 import traceback
 from collections import OrderedDict, defaultdict
+from itertools import chain
+from tqdm import tqdm
 
 from pymongo import MongoClient
 from pymongo import DESCENDING, ASCENDING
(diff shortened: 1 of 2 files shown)


computron:
Thanks!

This method is of course a bit more unsafe than usual since there is no rollback (every one in a while, I really miss SQL)

Can you give an idea of how much speedup there is for inserting 1000 workflows? 2X, 10X, 100X, etc.? I am not looking for an official test, just your general experience. I am asking because others have wanted bulk insertion speedup  (see https://github.com/materialsproject/fireworks/pull/240) so I would like to pull it, but just want to balance safety vs performance gain. I don't want to propose people sacrifice safety for a small performance boost for example.

computron:
Btw you can probably make the "unused variable n" go away using something like:

for _ in range(1000):

since "_" is usually recognized as a dummy/unused var

computron:
@richardjgowers this PR is relevant to you based on our previous discussions in case you have any comments / thoughts

richardjgowers:
@computron thanks for the ping, I'll have a look at this over the weekend

montoyjh:
Yep.  I'll benchmark and get back to you.

fireworks/core/tests/test_launchpad.py
```diff
@@ -113,6 +114,22 @@ def test_add_wf(self):
...
+            wfs.append(Workflow([Firework(ftask, name='lorem')
+                                 for n in range(0, randint(3, 10))],
+                                name='lorem workflow'))
+        self.lp.add_wfs(wfs)
+        num_fws_total = sum([len(wf.fws) for wf in wfs])
+        distinct_fw_ids = self.lp.fireworks.distinct('fw_id', {'name': 'lorem'})
+        self.assertEqual(len(distinct_fw_ids), num_fws_total)
+        num_wfs_in_db = len(self.lp.get_wf_ids({"name": "lorem workflow"}))
+        self.assertEqual(num_wfs_in_db, len(wfs))
+        self.lp.reset('', require_password=False, max_reset_wo_password=1000)
```
richardjgowers: This cleanup/teardown won't run if the test ever fails.  This test class already has a teardown method so should be fine without this

fireworks/core/tests/test_launchpad.py
```diff
@@ -113,6 +114,22 @@ def test_add_wf(self):
         self.assertEqual(len(fw_ids), 3)
         self.lp.reset('',require_password=False)
 
+    def test_add_wfs(self):
+        ftask = ScriptTask.from_str('echo "lorem ipsum"')
+        wfs = []
+        for n in range(100):
+            # create workflows with 3-10 simple fireworks
+            wfs.append(Workflow([Firework(ftask, name='lorem')
+                                 for n in range(0, randint(3, 10))],
```
richardjgowers: I'm not sure what the `randint` adds apart from making it harder to know how many Fws you've added, making the test a little harder to read

fireworks/core/launchpad.py
```diff
@@ -313,6 +315,47 @@ def add_wf(self, wf, reassign_all=True):
...
+
+            # Set root fws to READY
+            for fw_id in wf.root_fw_ids:
+                wf.id_fw[fw_id].state = 'READY'
+                wf.fw_states[fw_id] = 'READY'
+
+        # Insert all fws and wfs, do workflows first so fws don't
+        # get checked out prematurely
+        self.workflows.insert_many([wf.to_db_dict() for wf in wfs])
+        all_fws = chain.from_iterable([wf.fws for wf in wfs])
```
richardjgowers: If you want to chase performance when this scales to 1,000s, you'll want to use generator expressions not list comprehensions.  So `chain.from_iterable(wf.fws for wf in wfs)` without the square brackets

fireworks/core/launchpad.py
```diff
@@ -313,6 +315,47 @@ def add_wf(self, wf, reassign_all=True):
...
+        self.fireworks.insert_many([fw.to_db_dict() for fw in all_fws])
```
richardjgowers: Similarly, pymongo accepts generator expressions here too, so you can pass this as `(fw.to_db_dict() for fw in all_fws)` and allow pymongo to lazily evaluate the Fws

computron:
Thanks @richardjgowers for the helpful comments!

@montoyjh let me know when you have a revised version ready for re-review

fireworks/core/tests/test_launchpad.py
montoyjh: It doesn't work without the max_reset_wo_password argument, but I'll just move that to the teardown method.

fireworks/core/tests/test_launchpad.py
montoyjh: I want to ensure that sets of workflows of different firework lengths get tested (it's where I'd imagine the associated methods would fail), but I'll do this deliberately, rather than randomly.

montoyjh:
Ok, a new revision, here are some timings, seems like bulk addition gives about a factor of 10 speedup.  I'm really not 100% confident this is all that robust, so I'm reluctant to merge the tried and true `lpad.add_wf` functionality and this one (in fact renamed the function to `bulk_add_wfs` to suggest the bulk write operation and perhaps to discourage accidental use).  Up to you though.

fws_timings

computron:
Ok pulling as-is. It agree it would be better to have a single solution for both bulk and single insertion but I guess this is better than the current iteration of the code.

PR #263, opened 2018-03-09
Add support for pickle/unpickle of Firetasks

mlhenderson:
- Removes FiretaskMeta.\_\_call__, moves required parameter checks to FiretaskBase.\_\_init__
- Adds \_\_getstate__, \_\_setstate__, \_\_reduce__" to FiretaskBase
- Adds basic unit tests for pickle/unpickle to fireworks/core/tests/test_firework.py
- Small adjustment to one assert check in FiretaskBaseTest, since \_\_call__ was removed

The current behavior before this update throws a ValueError during unpickle because of how unpickle rebuilds the object, executing the metaclass \_\_call__ method with no arguments.

Originally, I had a workaround in FiretaskMeta that checked whether the inputs to \_\_call__ were length zero, and if so then bypass checking the required parameters.  This approach didn't cover all cases where you want to enforce required parameters, for example, ScriptTask().

Ultimately, it was more beneficial to move the parameter checks to FiretaskBase.\_\_init__, and then include a small number of magic methods to ensure pickling/unpickling could operate normally.

This *does* cause a small behavior change due to the removal of \_\_call__.  You can still instantiate a Firetask with the same syntax e.g., "cls()", you just no longer invoke metaclass.\_\_call__.

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -43,11 +43,7 @@ __email__ = "ajain@lbl.gov"
 __date__ = "Feb 5, 2013"
 
 
-class FiretaskMeta(abc.ABCMeta):
-    pass
-
-
-@add_metaclass(FiretaskMeta)
+@add_metaclass(abc.ABCMeta)
 class FiretaskBase(defaultdict, FWSerializable):
     """
     FiretaskBase is used like an abstract class that defines a computing task
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -44,12 +44,7 @@ __date__ = "Feb 5, 2013"
 
 
 class FiretaskMeta(abc.ABCMeta):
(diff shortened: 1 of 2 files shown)


computron:
Thanks!

Is there any need for FireTaskMeta anymore? Can we just remove it altogether?

mlhenderson:
FiretaskMeta is not doing anything now other than effectively applying abc.ABCMeta as the metaclass for FiretaskBase.  I didn't remove it entirely because I wasn't sure if you had other intentions for it.

I can remove the FiretaskMeta declaration and adjust the add_metaclass decorator for FiretaskBase.

computron:
I didnt have any other intentions for it, so removing it to avoid confusion
/ extra code would be great!

computron:
Thanks!

PR #264, opened 2018-03-10
Option to delete launch dirs when deleting workflows

gpetretto:
This PR introduces the possibility to delete all the launch directories associated with a workflow when calling delete_wf. I often use this option when deleting failed workflows, in order to avoid leaving large unreferenced temporary files in the file system. 
I thought that this may be of general interest.
If this is the case, two further modifications could be considered:
-  print the launch directories and add a check from the user before actually deleting them
- add warnings in case shutil.rmtree fails to delete the directory (with "onerror")

### DIFF
--- a/fireworks/core/launchpad.py
+++ b/fireworks/core/launchpad.py
@@ -12,6 +12,7 @@ import os
 import random
 import time
 import traceback
+import shutil
 from collections import OrderedDict, defaultdict
 from itertools import chain
 from tqdm import tqdm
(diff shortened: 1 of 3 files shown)


computron:
Thanks!

A couple of notes:

* I'd rename ``delete_ldirs`` to ``delete_launch_dirs``. It is longer but clearer
* You should probably(?) also delete all the archived launches. i.e., if someone has rerun a FW in the workflow, the previous launch will go into the archived_launches key. Sometimes a FW might have 5 archived_launches which likely also need cleanup?

computron:
Actually nevermind the second point,  I see that you've implemented the archived_launches already (didn't see it first since the PR review didn't show that portion of the code)

computron:
(to save time, I will just merge and then refactor the name)

PR #265, opened 2018-03-12
update fireworks' fw_ids in bulk insert

gpetretto:
While testing the new `bulk_add_wfs` function I realized that the fw_id are inserted with negative values in the fireworks collection. They should be updated explicitly also in the Fireworks instances before insert.


computron:
Thank you! 

@montoyjh

PR #303, opened 2018-08-02
Improve high dependency firework performance

tahorst:
For workflows with fireworks that have a lot of dependencies, the current code can take a long time.  Instead of getting the state of all parents and checking if all of those states are a completed state, we can check for the first parent that is not completed and exit early.  This isn't guaranteed to improve performance since the last parent in the list could be the only one that isn't complete but from what I've seen so far, the last parent to run/complete is usually the first checked in this list.

I have created a simple test script to show the difference in execution time [here](https://github.com/tahorst/fireworks/blob/master/test_high_dependencies.py) (but not included in this PR).  With my setup and 500 dependent tasks, the execution time between the task completing and the rocket finishing goes from 40 sec to ~1.5 sec.

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -982,16 +982,15 @@ class Workflow(FWSerializable):
             self.fw_states[fw_id] = fw.state
             return updated_ids
 
-        # what are the parent states?
-        parent_states = [self.id_fw[p].state for p in self.links.parent_links.get(fw_id, [])]
-
         completed_parent_states = ['COMPLETED']
         if fw.spec.get('_allow_fizzled_parents'):
             completed_parent_states.append('FIZZLED')
 
-        if len(parent_states) != 0 and not all(s in completed_parent_states for s in parent_states):
-            m_state = 'WAITING'
-
+        # check parent states for any that are not completed
+        for parent in self.links.parent_links.get(fw_id, []):
+            if self.id_fw[parent].state not in completed_parent_states:
+                m_state = 'WAITING'
+                break
         else:  # not DEFUSED/ARCHIVED, and all parents are done running. Now the state depends on the launch status
             # my state depends on launch whose state has the highest 'score' in STATE_RANKS
             m_launch = self._get_representative_launch(fw)


1fish2:
Nice!

Is it slow for making many remote DB queries? If so, could it make a single query then scan the results? I'm unfamiliar with MongoDB.

computron:
@tahorst thanks very much for this PR. I looked at it and confirmed that for the provided example it really speeds things up.

As you and @1fish2 mentioned, things might indeed not be quite as fast if the "unfinished" parent is not towards the beginning of the list. If it is toward the end of the list, you'd have similar performance as before.

With that in mind, I coded up a very related but different solution that takes advantage of the fact that the workflow refresh initially creates a Workflow object with "LazyFireWork" objects.

fireworks.core.launchpad.LaunchPad#get_wf_by_fw_id_lzyfw

The LazyFireWork objects only fetch data as needed - e.g., they only get the state when you need them to. Currently (before your PR), the code is fetching the state for every single parent FW - one at a time - in a "lazy" way. With your PR, the code fetches the state one at a time only until needs to, stopping when it hits an unfinished parent. So - better. However, with the new commit we bypass this altogether. This is because the #get_wf_by_fw_id_lzyfw method actually has a ``fw_states`` variable that is available to allow you to get the states of each Firework without doing any additional queries whatsoever. So, if we just use the ``fw_states`` variable to get the state rather than asking the LazyFirework to fetch the state, it should be as fast as your solution if the "unfinished" Firework is toward the beginning of the list, but also stay just as fast if the "unfinished" Firework is near the end.

The commit id is here:

	d9a71ac27cfd025af39f507137c407657e17dd38

and I will release a new FWS version on PyPI (1.7.6) that incorporates this change.

Although I am closing this PR without merging, I really want to thank you for getting us 90% of the way there and also for including the example that allowed me to really test all solutions. If you see any issues or improvements with this alternate solution, please let me know, but hopefully it should be just as fast or faster and serve your needs OK.

tahorst:
Looks great!  Thanks!

Issue #313, opened 2018-12-10
Problem with many detours


gpetretto:
I wanted to discuss about an issue connected to the maximum allowed size of the MongoDB document. 

If at some point in a workflow the generation of a large number of detours is needed, each with some specific input, the operation may fail with the error:
> pymongo.errors.DocumentTooLarge: command document too large. 

This is due to the fact that the whole `FWAction` (and thus all the detours with their inputs) is stored in the launch document.

Here is an example demonstrating the problem: https://gist.github.com/gpetretto/025d24c7a8f21c6ec40c6610b21424aa

One could avoid this kind of problem by generating the whole set of fireworks in the beginning, but this may not always be feasible and I have been asked to handle a case where each specific input is generated in the "parent" firework. 

Another workaround could be to save the inputs somewhere in the file system or in a database and let each child retrieve the required information. This does not seem so appealing, as it implies knowing beforehand whether this can happen or not and complicates the logic of the tasks.

So I would like to check if you would be interested in one of the following modifications (or any other) that would allow such a workflow to keep working irrespectively of the number of detours:
1) add an option in `fw_config` to completely disable the saving of the `FWAction` inside the launch document (or at least the `additions` and `detours` keys.)
2) remove the action from the launch if the size of the launch document exceeds the 16 MB limit:
```python
try:
    self.launches.find_one_and_replace({'launch_id': m_launch.launch_id},
                                       m_launch.to_db_dict(), upsert=True)
except DocumentTooLarge as e:
    # remove the action from the launch and try again
```
3) if the size of the launch document exceeds the 16 MB limit save the `FWAction` information to GridFS and store its id in place of the action in the launch document.

As a side note, of course the same problem will show up if a large number of fireworks try to update/mod the spec of a common child FW with a large enough amount of data. However this would not be solved so easily at the fireworks' level and seems easier to avoid with small modifications to the design of the wf.

Thanks

computron:
Hi @gpetretto ,

Thanks very much for reporting this, for offering potential solutions as well as an offer to implement.

I think most users will probably prefer option 3. This will require no intervention from the user, retain all the information, and keep things working "normally" for most users.

You could have the name of the fallback GridFS collection be an option in the FWConfig (with some default value) if desired. If the user purposely sets this value to None/null, then the operation would essentially follow your second solution, i.e., the information is removed (perhaps instead of removing entirely, replace some string description like "5 FWS were added; information removed due to document size restrictions").

Would this work?

PR #314, opened 2019-01-29
[WIP] Solve DocumentTooLarge error for Launch

gpetretto:
This PR addresses the problem discussed in #313.
At this stage I have implemented only the saving to gridfs. While I would have liked to include the option to switch to the deletion of the data, I came across a problem that I did not consider beforehand:  
at present, the call to `Workflow.apply_action` is performed inside `Workflow.refresh` and the Launch object used there is not the one available in `complete_launch`, so it is downloaded from the DB there. With this configuration, if the action is not stored anywhere it will not be applied. 

A possible workaround would be to make the launch an argument of `refresh` and propagate the original object from `complete_launch` (through `Launchpad._refresh_wf`), but I am not sure if it would be a good idea to do that just to allow this option.
Do you have any suggestion about how to avoid this problem in another way?

A couple of further notes on the current implementation: 
* I added a new attribute to the `Launch` to store the potential gridfs id instead of setting it in place of the `action`. This is to avoid potential serialization/deserialization problems if the content of the `action` attribute is not an `FWAction`.
* I purposedly tried to be general in the naming of this gridfs functionality instead of tying it explicitly to the action in Launch. So that, in case some other cases of `DocumentTooLarge` errors are identified, they could be dealt with in a similar manner and use the same collection. Is this fine or would you prefer to have different collections in case some other properties requiring gridfs will show up?

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -393,7 +393,7 @@ class Launch(FWSerializable, object):
     """
 
     def __init__(self, state, launch_dir, fworker=None, host=None, ip=None, trackers=None,
-                 action=None, action_gridfs_id=None, state_history=None, launch_id=None, fw_id=None):
+                 action=None, state_history=None, launch_id=None, fw_id=None):
         """
         Args:
             state (str): the state of the Launch (e.g. RUNNING, COMPLETED)
@@ -403,8 +403,6 @@ class Launch(FWSerializable, object):
             ip (str): the IP address where the launch took place (set automatically if None)
             trackers ([Tracker]): File Trackers for this Launch
             action (FWAction): the output of the Launch
-            action_gridfs_id (str): the document id of the gridfs file where the action is
-                stored in case the size of the Launch document exceeds the 16MB limit.
             state_history ([dict]): a history of all states of the Launch and when they occurred
             launch_id (int): launch_id set by the LaunchPad
             fw_id (int): id of the Firework this Launch is running
@@ -417,7 +415,6 @@ class Launch(FWSerializable, object):
         self.ip = ip or get_my_ip()
         self.trackers = trackers if trackers else []
         self.action = action if action else None
-        self.action_gridfs_id = action_gridfs_id
         self.state_history = state_history if state_history else []
         self.state = state
         self.launch_id = launch_id
@@ -530,7 +527,6 @@ class Launch(FWSerializable, object):
                 'ip': self.ip,
                 'trackers': self.trackers,
                 'action': self.action,
-                'action_gridfs_id': self.action_gridfs_id,
                 'state': self.state,
                 'state_history': self.state_history,
                 'launch_id': self.launch_id}
@@ -552,7 +548,7 @@ class Launch(FWSerializable, object):
         action = FWAction.from_dict(m_dict['action']) if m_dict.get('action') else None
         trackers = [Tracker.from_dict(f) for f in m_dict['trackers']] if m_dict.get('trackers') else None
         return Launch(m_dict['state'], m_dict['launch_dir'], fworker,
-                      m_dict['host'], m_dict['ip'], trackers, action, m_dict["action_gridfs_id"],
+                      m_dict['host'], m_dict['ip'], trackers, action,
                       m_dict['state_history'], m_dict['launch_id'], m_dict['fw_id'])
 
     def _update_state_history(self, state):
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -393,7 +393,7 @@ class Launch(FWSerializable, object):
     """
 
     def __init__(self, state, launch_dir, fworker=None, host=None, ip=None, trackers=None,
-                 action=None, state_history=None, launch_id=None, fw_id=None):
+                 action=None, action_gridfs_id=None, state_history=None, launch_id=None, fw_id=None):
         """
         Args:
             state (str): the state of the Launch (e.g. RUNNING, COMPLETED)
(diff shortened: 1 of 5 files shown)


fireworks/core/tests/tasks.py
```diff
@@ -74,4 +74,22 @@ def run_task(self, fw_spec):
...
+
+@explicit_serialize
+class DetoursTask(FiretaskBase):
+    optional_params = ["n_detours", "data_per_detour"]
+
+    def run_task(self, fw_spec):
+        data_per_detour = self.get("data_per_detour", None)
+        n_detours = self.get("n_detours", 10)
+        fws = []
+        for i in range(n_detours):
```
computron: since the variable ```i``` is unused, better to call it ```_``` per Python convention

fireworks/core/launchpad.py
```diff
@@ -1269,8 +1291,28 @@ def complete_launch(self, launch_id, action=None, state='COMPLETED'):
...
 
-        self.launches.find_one_and_replace({'launch_id': m_launch.launch_id}, m_launch.to_db_dict(),
-                                           upsert=True)
+        try:
+            self.launches.find_one_and_replace({'launch_id': m_launch.launch_id},
+                                               m_launch.to_db_dict(), upsert=True)
+        except DocumentTooLarge:
+            launch_db_dict = m_launch.to_db_dict()
+            action_dict = launch_db_dict.get("action", None)
+            if self.gridfs_fallback is None or not action_dict:
```
computron: these seem like 2 different errors

If there's no action_dict, reraising the error seems appropriate

If there is an action_dict but the user has not set gridfs_fallback, perhaps you should add to the error that the user should set a gridfs fallback collection

fireworks/core/firework.py
```diff
@@ -393,7 +393,7 @@ class Launch(FWSerializable, object):
     """
 
     def __init__(self, state, launch_dir, fworker=None, host=None, ip=None, trackers=None,
-                 action=None, state_history=None, launch_id=None, fw_id=None):
+                 action=None, action_gridfs_id=None, state_history=None, launch_id=None, fw_id=None):
```
computron: I would prefer it if the core Launch object itself did not have an action_gridfs_id attribute. It's better if the Launch object can not contain too many details of the specific database implementation that will be used to store it.

Is there a way to structure this so that the existing "action" attribute can easily distinguish whether it's a normal action or a gridfs id? e.g., perhaps if the action looks like this: action={"gridfs_id": "adsfasdf"} then the LaunchPad will know that the action needs retrieval from GridFS.

fireworks/core/launchpad.py
computron: (or at least raise a warning that the gridfs fallback was not set, but the document is too large)

fireworks/core/launchpad.py
```diff
@@ -397,6 +407,7 @@ def get_launch_by_id(self, launch_id):
         """
         m_launch = self.launches.find_one({'launch_id': launch_id})
         if m_launch:
+            m_launch = set_action_from_gridfs(m_launch, self.gridfs_fallback)
```
computron: Could you refactor the name:
``set_action_from_gridfs``

to the name:
``get_action_from_gridfs``

I think the code would also be clearer if there was some kind if-else outside the function. e.g., more like
```
if "gridfs_id" in m_launch.action:
    m_launch.action = get_action_from_gridfs(m_launch.action["gridfs_id"], self.gridfs_fallback)
```
This makes it clear to the developer that they don't need to worry about get_action_from_gridfs internals at all unless there is some weird gridfs thing going on and they can mentally skip over this line of code.

computron: Also in the example above, the function itself is not setting anything - it's just getting information

fireworks/core/firework.py
gpetretto: My point in adding one more attribute was to avoid potential problems while deserializing the Launch object (now the dictionary in the db may not contain an FWAction), but I see your point in wanting to preserve the Launch interface.  
I have implemented the modification that you suggested. As long as the Launch is generated or retrieved using theLaunchPad it should be safe.

fireworks/core/launchpad.py
gpetretto: I was seeing the things from another perspective: the `set_action_from_gridfs` function was (1) checking if the data should be retrieved from gridfs and (2) directly inserting it in the launch dictionary (thus the "set" in the name).  
If a developer has to work on the launch collection in some other point of the code, now he should be aware of the details about how the `gridfs_id` key is stored and repeat the check once more.  

I have implemented this modification following your suggestion anyway, so that we can have a better feeling of how it looks like.

computron: I see your point here; the new changes also result in a lot of "if statements" like this:
```
            if m_launch.get("action") and "gridfs_id" in m_launch["action"]:
```

How about something like:

```
for l in launches:
    l["action"] = get_action_from_gridfs(l["action"], self.gridfs_fallback)
```

with the ``if`` statement within the function itself.

I think the only change from your original in that case is that the getter code is simply retrieving the action rather than updating the entire launch object on your behalf. This modification just makes it more clear to me that (i) the only part of the launch that is being modified and (ii) that you are merely trying to retrieve existing information, not "setting" any new values. When I think of "setting" the action, I think that the code is somehow receiving *new* information and modifying the information accordingly (e.g., you set the action when you return a FWAction) rather than simply grabbing existing information from a database.

gpetretto: Thanks for all the comments. I updated the code according to your suggestions.

computron:
Thanks!

PR #315, opened 2019-02-11
Fixed jinfa functionality like include that involves finding other te…

water-e:
The original code used Template's constructor that implicitly sets up a jijja2 Environment that doesn't understand the file system, so it couldn't do things like include another template even if that template was side-by-side.


computron:
Thanks!

PR #316, opened 2019-02-13
Further authentication issues in LaunchPad construction

dyllamt:
# Summary

After the last changes to `LaunchPad`, users were always authenticating against the admin database. Therefore, they could not access databases that they were users of. These changes ensure that a user is authenticating against the database that they want to access, and not the admin database by default.

* A *not_the_admin_db* database is created by circle ci with a single user *myuser*
* The `authSource` is always passed into the `MongoClient` constructor
* Tests ensure that *myuser* can only authenticate against the database that they are a part of
* Tests ensure that the `authSource` is inferred from the database that the user wants access to

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ jobs:
       - run:
           name: Make tox env and setup mongo
           command: |
+            mongo not_the_admin_db --eval "db.createUser({user: \"myuser\", pwd: \"mypassword\", roles: [ \"dbOwner\" ]})"
             mkdir ~/tox_env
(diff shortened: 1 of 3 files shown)


dyllamt:
# Issues

CircleCI only uses the primary (first) docker image when running commands, which means that the `mongo` command is not available for creating the additional database and user for tests, which causes the tests to fail. At this point, does anyone have any suggestions for how we can get around this problem?

dyllamt:
you can create database new users through the pymongo drivers, so all the tests pass now.

computron:
Great thanks!

PR #317, opened 2019-02-13
corrected keyword 'function'->'func' and outputs string -> list

ikondov:
Recently I discovered keyword and type inconsistencies in two examples in the upstream/master branch. These inconsistencies seem to be leftovers after refactoring the PyTask before a merge some time ago. The relevant tests are not affected of course.


computron:
Thank you!

PR #318, opened 2019-03-05
Offline LaunchPad

richardjgowers:
Adds an OfflineLaunchPad class, which has a sqlite3 backend.  This isn't to replace LaunchPad, but just adds a different storage medium.  This hasn't modified anything outside of the LaunchPad, (so far).  The only point of grief I've found so far is that `FWorker.query` is hardcoded to be mongodb syntax.  Oh and `LaunchPad` methods take mongodb queries all over the show, this won't work with a sqlite backend.

Currently creating a Workflow, submitting it, and running it with `rocket_launch` works, so this kind of works?  All sorts of corner cases don't work yet....

ping @shyamd

### DIFF
--- a/fireworks/tests/mongo_tests.py
+++ b/fireworks/tests/mongo_tests.py
@@ -639,7 +639,11 @@ class SQLiteTests(MongoTests):
             n = len(c.execute('SELECT * FROM {}'.format(name)).fetchall())
         return n
 
+    def test_basic_fw_offline(self):
+        pass
 
+    def test_offline_fw_passinfo(self):
+        pass
 
 
 if __name__ == "__main__":
--- a/fireworks/tests/mongo_tests.py
+++ b/fireworks/tests/mongo_tests.py
@@ -40,10 +40,10 @@ MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
 NCORES_PARALLEL_TEST = 4
 
 def random_launch(lp_creds):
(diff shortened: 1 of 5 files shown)


shyamd:
Is it just user queries in mongodb syntax that are the issue or does are there hardcoded mongodb queries that cause issues with your sqlite implementation?

richardjgowers:
https://github.com/materialsproject/fireworks/blob/master/fireworks/core/fworker.py#L60

So far just this bit of FWorker. I think I can hack around it. Tbh going better than I had expected.

shyamd:
I would recommend you switch to unittest since the CI in Fireworks depends on nose.

fireworks/tests/mongo_tests.py
```diff
@@ -597,5 +607,44 @@ def test_stats(self):
         self.assertEqual((workflow_results[0]["_id"], workflow_results[0]["count"]), ("COMPLETED", 3))
 
 
+class SQLiteTests(MongoTests):
```
richardjgowers: @shyamd yeah I'm going through the existing mongo tests with this class, one by one.  unittest is pretty awful, the cleanup doesn't always seem to work so once one test fails they all do...

computron:
@richardjgowers I generally close PRs into master unless there has been activity in the last month. I am closing this for now, please feel free to re-open at a later time as needed.

PR #320, opened 2019-04-02
small updates

gpetretto:
This PR contains two modifications.

The first is a fix for a minor problem that I have ancountered in ``rerun_fw`` in combination with the ``recover_launch`` option. The ``recover_launch`` dictionary is coming from the ``checkpoint`` keyword in the Launch history. This can contain the serialized version of an MSONable object. This gets deserialized while loading, but when doing the
```python
self.fireworks.find_one_and_update({"fw_id": fw_id}, set_spec)
```
it is not serialized and the update of a dictionary containing this object fails. I added a ``recursive_dict`` that solves the problem and should be harmless for all the other cases. 

I just have the doubt if this should instead be done inside ``get_recovery``. The API of that method is not really explicit about what should be its output.

The second point is the addition of a ``-wf`` option for the ``set_priority`` command in ``lpad``. I sometimes want to set a priority for a whole workflow and this seemed a functionality that could be useful for the generic user.


computron:
Thanks!

PR #329, opened 2019-06-14
fix -t option in lpad get_wflows

gpetretto:
Not sure if it is really used, but currently the `-t` option for `lpad get_wflows` fails in case of 1 or 0 workflows listed


computron:
Thanks!

PR #330, opened 2019-06-26
Load according to server scheme (https or http)

dwinston:
Browsers like Firefox refuse to load files over http if the host page is served over https.


computron:
Thanks! Didn't know about that syntax

PR #336, opened 2019-08-07
Allow binary files in filepad

jotelha:
So far, it has only been possible to store text files in the filepad. Trying to store binary files threw an error. This simple modification stores all files as binary in the filepad. However, the filepad behavior would change in such a way that the responsibility of correctly encoding / decoding text files is passed on to the user.


computron:
Thanks! I will merge but then add a flag to go back to "plain" mode if needed, e.g., if someone is working between Windows / Linux and needs to be able to write in file mode

PR #337, opened 2019-08-07
_files_in from an arbitrary number of merging workflow branches

jotelha:
Use mod_spec instead of update_spec to pass on out-/infiles along Fireworks. 

That allows for files to come from several merging branches of a workflow as in
image
as long as they are labeled differently.

It seemed to me that with the previous "update_spec" utilization, only (a set of) _files_out from one branch were passed onto the child firework, depending on which branch executed last.


computron:
Thanks!

PR #345, opened 2019-09-09
Switch to Pytest for testing

shyuep:
I have left py2 and py3 test in. But fireworks should consider switching to py3 completely since py2 will be dead on Jan 1 2020. @computron should review. Tests pass on both travis and circle.


shyuep:
@computron This is ready for merge.

computron:
Thanks!

PR #347, opened 2019-09-19
fix for lpad

gpetretto:
I have found two potential problems related to the LaunchPad.

The first is in the `lpad` command. Since the `-l` option is assigned a default in argparse the `-c` is never taken into account. I removed the default and modified the `parse_helper` function so that it should properly consider the `-c` option as well.

The second problem concerns the fetching of the Launches from the database. 
Currently the launches are fetched in this way:
```python
list(self.launches.find({'launch_id': {"$in": fw_dict['launches']}}))
```
My understanding of how MongoDB works is that this query does not guarantee the order of the launches (and archived launches as well). This list is then passed directly to the Firework `__init__`. The incorrect order might be an issue, since in some portions of the code the last launch is accessed with `fw.launches[-1]`. e.g.
https://github.com/materialsproject/fireworks/blob/master/fireworks/core/launchpad.py#L1527
I propose to add a sorting when the launches are pulled from the DB. 
Do you agree that this is actually an issue? Would this solution work for you?

### DIFF
--- a/fireworks/core/launchpad.py
+++ b/fireworks/core/launchpad.py
@@ -444,11 +444,13 @@ class LaunchPad(FWSerializable):
         if not fw_dict:
             raise ValueError('No Firework exists with id: {}'.format(fw_id))
         # recreate launches from the launch collection
-        launches = list(self.launches.find({'launch_id': {"$in": fw_dict['launches']}}))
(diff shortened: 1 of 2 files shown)


computron:
Thanks!

PR #357, opened 2019-10-08
Update Google Groups references to Discourse

utf:
The FireWorks support forum is moving from Google Groups to Discourse (https://hackingmaterials.discourse.group/c/fireworks). I've updated all references to Google Groups to point to the new forum.

However, as we are not migrating old posts to the new forum, I've left in links that point to specific  Google Groups posts. Specifically, these are:

```
./fireworks/core/launchpad.py:                # see: https://groups.google.com/forum/#!topic/fireworkflows/oimFmE5tZ4E
./docs_rst/installation.rst:.. note:: We suggest that you use Python 2.7.3 or higher, especially in production. There is a `bug <https://groups.google.com/forum/#!topic/modwsgi/DW-SlIb07rE>`_ in Python 2.7.2 that could affect FireWorks (although we haven't seen any problems yet). As of FireWorks v0.7, Python 3.3 and higher should also work.
``` 

Once merged, the docs should be regenerated to reflect the updates.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,5 +1,5 @@
 Please submit help issues to:
-https://groups.google.com/forum/#!forum/fireworkflows
+https://hackingmaterials.discourse.group/c/fireworks
 
 The Github issues is no longer used except for internal development purposes.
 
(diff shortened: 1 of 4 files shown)


mkhorton:
This needs to be changed, the PR was stale

computron:
Well one step in the right direciton

What is the actual updated link?

computron:
nevermind found it:

https://discuss.matsci.org/c/fireworks

will update shortly

mkhorton:
Yep, that's it. We got the old Google groups discussions imported too :)

PR #365, opened 2019-11-07
Authentication from external sources

ikondov:
MongoDB supports external authentication mechanisms (x509 certificates, LDAP, Kerberos token) instead of internal username/password credentials. To enable them in FireWorks the ``authmechanism`` keyword has to be added to the LaunchPad object. This branch enables this feature and has been tested with x509 CA-signed certificates. The documentation is extended accordingly. LDAP, Kerberos and other external methods have not been tested.


computron:
Thanks! Note that after this one I think I am just going to add a generic way to add kwargs to LaunchPad objects rather than supporting each kwarg one at a time

computron:
actually I just removed this parameter in favor of using the mongoclient_kwargs, as there was a problem with the default not being set in this PR

PR #379, opened 2019-12-04
avoid checking the number of jobs in the queue if not needed

gpetretto:
Sometimes the queuing system is not responsive and when running `qlaunch`the check for the number of jobs goes in timeout. Two improvements implemented here to avoid this issue:
* if the maximum jobs in the queue is 0 (no limit), do not check the current number of jobs. 
* added a timeout option to the `CommonAdapter`. Before the timeout was hardcoded to 5 seconds.


computron:
Thanks!

PR #380, opened 2019-12-05
Avoid failures due to encoding issues in tracker

jotelha:
Tracking files in text mode used to cause Fireworks to fail if encoding issues arose.
This pull request or specifying a particular encoding, i.e. "utf-8" should avoid such failures.


computron:
Thanks!

PR #381, opened 2019-12-06
params-fix

mamachra:
Currently, the FireTasks as implemented in fileio_tasks are missing some required and optional parameters. 
Adequate tags are added in order to pass the consistency check upon initialization of the FiretaskBase.


computron:
Thanks!

PR #384, opened 2019-12-11
Extended get files by query task

jotelha:
A suggestion:

Turns the previously contributed "GetFilesByQueryTask" into something useful:

-  Allows for nested queries to be stored by using "->" instead of "." separator.
-  Different options to handle a priori uncertain number and names of queried files (see task's doc string)
 - Tests for these added features (also serving documentation purposes)

Other additions:

- Use with-statement to enforce closing all file streams in other FilePad-Task tests, avoid warnings
- Tiny helper function "arrow_to_dot" within "utilities/dict_mods.py" facilitates "->" to "." conversion.

Best regards,

Johannes

### DIFF
--- a/fireworks/user_objects/firetasks/filepad_tasks.py
+++ b/fireworks/user_objects/firetasks/filepad_tasks.py
@@ -108,6 +108,8 @@ class GetFilesByQueryTask(FiretaskBase):
           directory.
         - new_file_names ([str]): if provided, the retrieved files will be
           renamed. Not recommended as order and number of queried files not fixed.
+        - meta_file (bool): default: False. If True, then metadata of queried files written to
+          a .yaml file of same name as file itself, suffixed by...
         - meta_file_suffix (str): if not None, metadata for each file is written
           to a YAML file of the same name, suffixed by this string.
           Default: ".meta.yaml"
@@ -134,6 +136,7 @@ class GetFilesByQueryTask(FiretaskBase):
         fizzle_empty_result         = self.get("fizzle_empty_result", True)
         fizzle_degenerate_file_name = self.get("fizzle_degenerate_file_name",
                                                 True)
+        meta_file                   = self.get("meta_file",False)
         meta_file_suffix            = self.get("meta_file_suffix",".meta.yaml")
 
         assert isinstance(query,dict)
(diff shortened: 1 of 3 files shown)


jotelha:
All tests fail here and the "ci/circleci: py3test" "Details" link leads to 

https://circleci.com/gh/materialsproject/fireworks/2387?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

with message "Project not found". Something broken? Locally, the added tests ran without issues using the python-3.8.8 unittest framework.

computron:
Thanks! Can you clarify the reason for switching from dot to arrow? You wrote "allows for nested queries to be stored" but I don't think I understood why a nested query can't be stored if there is a "." in it.

jotelha:
Queries are expected as nested dicts and not as plain strings, thus the -> aliasing for the . (dot) separator allows to store queries like this:

```Python
ft = GetFilesByQueryTask(
    query={
        'metadata->project':    project_id,
        'metadata->type':       'surfactant_file',
    },
    sort_key='metadata.datetime',
    sort_direction=pymongo.DESCENDING,
    limit=1,
    new_file_names=['default.pdb'])]
```

If I remember correctly, the MongoDB language does not allow for dots in keys. It's the same issue as in the dict_mods.py file at https://github.com/materialsproject/fireworks/blob/07bace776fedefd09907272334a2c5925ffce51d/fireworks/utilities/dict_mods.py#L55-L58.

computron:
Ah yes, I remember - MongoDB doesn't allow storing dictionaries where the keys have a dot in them. So storing the parameter:

```{"query": {"key.subkey": "value"}}```

can't be done - making it difficult to serialize the FireTask. The arrows should indeed make it possible to store the query and thereby serialize the Firetask. Merging this now along with the other improvements, thanks!

jotelha:
Exactly. A related note: Similarly, I believe, it is not possible to store any query involving $-prefixed operators, i.e. 
    
    {'metadata.datetime': {'$gt': '2020'} }

thus it might be a good idea to store queries as plain strings instead. Are there any mongo-language-specific serialization recommendations for query documents?

computron:
I don't know of any mongo language specific serialization recommendations; it's possible that a simple string is best.

As an aside, it looks like as of MongoDB 3.6+, dots are allowed in key names. But dollar sign prefixes are still prohibited:

https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names

PR #393, opened 2020-03-12
Add support for dataclass Firetasks

utf:
## Summary

This pull request extends the Firetasks constructor to support python dataclasses. This allows Firetasks to be explorable when using jupyter notebooks (i.e., tab completion), to support type hinting, and also to have explicit default values.

The changes implemented in this pull request are backwards compatible and will not affect existing Fireworks or their operation.

I believe this approach will make documenting and troubleshooting Firetasks more reliable. 

### Initialisation

Firetasks can now be initialised using a dataclass approach for specifying required and optional arguments.

```python
@dataclass(repr=False)
class Task(FiretaskBase):
    param: str
    param_a: int = 0

task = Task(param=123)
```

Is equivalent to:
```python
class Task(FiretaskBase):
    required_params = ["param"]
    optional_params = ["param_a"]

task = Task(param=123)
```

In the dataclass case we get type hinting and default arguments for free. 

### Attribute access

Firetask attributes can now be accessed using a dot syntax whether or not they were created with the dataclass syntax. Furthermore, the traditional dictionary style access for attributes also works in both cases. For example, all of these are valid:

```python
task.param = 1
task["param_a"] = 2
task["param"]
task.get("param_a", 123)
```

## Notes

- The `repr=False` syntax is not strictly necessary but means the string representation for both fireworks will be identical.
- I'm doing additional testing to ensure the new FiretaskBase implementation does not break anything in atomate.
- Casting the Firetask to a dictionary directly using `dict(task)` does not perform the same for both implementations (the dataclass method will return an empty dict). From what I can tell this is not a recommended approach anyway and `to_dict` should always be used. If this is a deal breaker I can try and fix this behaviour.
- The use of dataclasses requires python 3.6+. Considering fireworks is already running CircleCI only for python 3.7+ I don't think this should be too much of an issue.

I am open for suggestions on how this PR can be improved and whether there are any obvious holes in this approach.

This PR came out of the discussion in this thread hackingmaterials/atomate#377.


mkhorton:
You could probably override `__iter__` if you want to replicate the old dict behavior, dataclasses have an `asdict()` method (I'm not sure if there was another issue here? agreed not very important anyway).

mkhorton:
Something like this should do it:

```
@dataclass(repr=False)
class Task(FiretaskBase):
    param: str
    param_a: int = 0

    def __iter__(self):
        return iter(asdict(self).items())

task = Task(param=123)
dict(task)  # returns {'param': 123, 'param_a': 0}
```

utf:
Unfortunately, the `__iter__` method will be ignored if the class is already a dict subclass. See: https://stackoverflow.com/questions/32250380/hook-type-casting-to-dict-in-python

One solution would be to remove the defaultdict inheritance and use an internal private dictionary to store the args and kwargs for the non dataclass case. You could then use the `__iter__` method like you suggested.

mkhorton:
> But I'm not sure it is worth it unless the dict casting is a major issue.

Agreed. Thanks for the link though, interesting!

computron:
@utf can you let me know the latest status of this PR?

computron:
closing this PR due to inactivity, re-open if needed

mkhorton:
What needs to be done here? It looks ready to merge

computron:
I asked Alex awhile ago if the PR was ready and he said he was still working on it ... anyway please re-open and say it's ready for review if that is the case.

mkhorton:
Alright, will defer to @utf on that, looks like a nice feature to add though -- the main benefit is the type checking that comes in for free with the data classes imo.

PR #394, opened 2020-04-01
JSON schema

ikondov:
This branch implements

- JSON schema for the most serializable objects in FireWorks in the `schema` folder
- a schema validator in the one of the tests `fireworks/utilities/tests/test_json_schema.py`
- a custom instance validator in `fireworks/utilities/json_schema.py`
- direct validator tests `fireworks/utilities/tests/test_json_schema.py` using a set of samples in `fireworks/utilities/tests/schema_samples`
- automatic validation in `from_file()` and in `load_object_from_file` is possible but must be turned on by the user by setting `JSON_SCHEMA_VALIDATE` to `true` (what objects get validated can be controlled via `JSON_SCHEMA_VALIDATE_LIST`)
- further tests invoking the instance validator via `from_file()` in `fireworks/utilities/tests/test_json_schema.py`
- short tutorial in `json_schema.rst`

For a background discussion see https://discuss.matsci.org/t/json-schema-for-fireworks/3655.


ikondov:
Followed the "quick" strategy, the time stamp validation error does not occur. Unfortunately, the problem with deserialization before schema validation is persisting: adding a sample with a `_dupefinder` fails. See the test output. The deserialization of `DupeFinderExact` is carried out before JSON schema validation begins. Note that the input is valid because the direct validation test is passed. The same happens with `BackgroundTask` in a sample with `_background_tasks`. This means that if validation is enforced in all `from_dict()` methods the code will crash with valid inputs that contain these spec keys.

Lets us omit validating dictionaries (if produced by `from_dict()` they are valid anyway). A good place to validate is right after the input is loaded from file. For all objects this is in `fireworks/utilities/fw_serializers.py` in the `from_format()` and `load_object_from_file()`. But there we do not know the name of the calling class that we need for calling the validator with the relevant scheme. I will try to find out how this can be done. Other suggestions are welcome.

ikondov:
In the head revision the following has changed:
1. Removed the validation calls *only* from `from_dict()` methods that are decorated with `recursive_deserialize()`. Validation for these classes will be only done for inputs from file.
2. Added validation to `from_format()` class methods and to the `load_object_from_file()` function.

One thing can be done is to remove validation calls from *all* `from_dict()` methods: in this case doubled validation (due to `from_format()` -> `from_dict()` call) will be avoided but there will be no validation of dictionary inputs for any classes.

ikondov:
Only small changes addressing the potential issues mentioned above.

computron:
Hi,

First of all, thanks for your efforts in creating this initiative.

I have two concerns about the JSON validation:

1. If we always validate on every deserialize, this is going to add time to workflows. Some FireWorks users like to run many (e.g., millions) of small Fireworks. While the FireWorks software already has many time overheads that make this kind of situation not an ideal case of Fireworks, I am not sure we want to add even *more* overhead time to the usage of Fireworks.

2. This does add a little more coding effort when creating a new subclass of FWSerializable, since one needs to now also add a new schema JSON otherwise validation will fail. While I am not against adding another step needed to subclass FWSerializable, I'd need to know that the gains outweigh the downside. There are also complications in that there are potentially downstream packages from FireWorks that may also want to introduce new objects that are FWSerializable.  Those downstream packages will not be able to add a schema into the main FireWorks package. Furthermore, even small changes like default value modifications will mean we are supposed to go in and modify the schema to match - essentially having to code everything twice. Generally, there are enough moving parts and "gotchas" with FireWorks that I hesitate to add more layers of complication.

Could you let me know a little more about why you want to add this feature, and especially validate on every deserialize? In practice, I personally have not really run into issues here. So, I'd like to know what is the potential gain and what issues you have run up against so it can be weighed against the downsides.

ikondov:
Thank you for your feedback. I agree with your two points. Below you will find my explanations.

>     1. If we always validate on every deserialize, this is going to add time to workflows. Some FireWorks users like to run many (e.g., millions) of small Fireworks. While the FireWorks software already has many time overheads that make this kind of situation not an ideal case of Fireworks, I am not sure we want to add even _more_ overhead time to the usage of Fireworks.

Validation might be a performance issue for large documents or a large number of input files. But I have made no scaling tests yet to see whether validation is a scaling bottleneck or some other overhead that you mention is dominant. I will do some scaling tests and communicate the result.

>     2. This does add a little more coding effort when creating a new subclass of FWSerializable, since one needs to now also add a new schema JSON otherwise validation will fail. While I am not against adding another step needed to subclass FWSerializable, I'd need to know that the gains outweigh the downside. There are also complications in that there are potentially downstream packages from FireWorks that may also want to introduce new objects that are FWSerializable.  Those downstream packages will not be able to add a schema into the main FireWorks package. Furthermore, even small changes like default value modifications will mean we are supposed to go in and modify the schema to match - essentially having to code everything twice. Generally, there are enough moving parts and "gotchas" with FireWorks that I hesitate to add more layers of complication.

My working assumption was that only the built-in FWSerializable subclasses and a few further user subclasses defined in the FireWorks package will be affected. To be honest, I did not consider use of FWSerializable in downstream packages. For me, the solution is to enforce validation only to built-in subclasses. I have to find out how this will technically be implemented but please first let me know whether this is an acceptable suggestion.

> Could you let me know a little more about why you want to add this feature, and especially validate on every deserialize? In practice, I personally have not really run into issues here. So, I'd like to know what is the potential gain and what issues you have run up against so it can be weighed against the downsides.

Regarding JSON schema alone: In my group we work on a workflow editor that will benefit from a machine-readable formal description (->domain-specific language) of the JSON/YAML data used in FireWorks. In addition, we work on interoperability between different workflow systems and view JSON schema as a good start for translator design.

Regarding JSON schema validation: FireWorks users who use JSON/YAML as input format will benefit due to early detection of input data that are incompatible with FireWorks. Such invalid inputs result later in runtime errors that are more difficult to handle by the user and sometimes in wasted computing resources.

computron:
I think my overall thoughts are:

Defining a JSON schema for documentation / external translators - certainly, this would be nice to have. I worry a little about whether it will be kept up to date properly in case there are any changes. But, there hasn't been much change to the schema for quite awhile, so maybe this is not such a practical concern.

Defining a JSON schema for validation - again this more where I have hesitation. If we only validate the core FireWorks objects (and not user subclasses) I worry that we'll be doing a lot of checks (e.g., a million checks involving a file opening of a schema file / etc) which will slow down the system. Further, adding any additional layer creates more code to maintain and potential to break things (perhaps unnecessarily) for people. Perhaps one way to get some kind of middle-ground would be to run the validators when loading from_file() but not from_dict(). That way, anyone trying to convert JSON/YAML to an object would have the validator run. But, the internal operation of FireWorks would not be subject to the constant validation.

ikondov:
Following your suggestions: 
 - The are no validator calls in `from_dict()` methods.
 - Validation in `from_file()` is done only if configuration variable `JSON_SCHEMA_VALIDATE` is `True`. The default value is `False`. This means there will be no impact on performance unless the user turns the value to `True` in their `FW_config.yaml`.
 - Validation is done only if the class name is listed in `JSON_SCHEMA_VALIDATE_LIST`. The default list includes only names of objects that are often read from files.

Some of the tests are skipped if `JSON_SCHEMA_VALIDATE` is `False`. Three tests are always run to keep the new code (only 41 lines) covered for the CI. For this jsonschema package should stay in `requirements.txt`.

There is no additional layer that would be hard to maintain. In `fw_serializers.py` this is just 6 lines difference after the change. I will maintain the new files.

If you accept this revision I can add some docs about this topic.

computron:
Hi,

Thanks again for making these changes. I had a chance to think them over and wanted to suggest moving some of the Python code and all of the JSON files to an external package. The reason is because I don't think we'll be able to support upkeep of the JSON schema files ourselves, and any issues with those JSON files should probably be maintained elsewhere than core FireWorks. Here is specifically what I suggest:

1. Create a new package called ```fws_json_schema``` (or whatever you want to name) it in your own Github account.
2. Move the JSON files as well as ```fw_json_schema.py``` to the new repo.
3. Move the setting ```JSON_SCHEMA_VALIDATE_LIST``` to your new package. (Keep the parameter ```JSON_SCHEMA_VALIDATE``` in the main FWS repo). 
4. In the main FireWorks repo, for the modules that use JSON_SCHEMA (looks like just ```fw_serializers.py```, have a code section at the top of the import section that says:
```
if JSON_SCHEMA_VALIDATE:
    import fw_json_schema.json_schema as json_schema
    JSON_SCHEMA_VALIDATION_LIST = fw_json_schema.JSON_SCHEMA_VALIDATION_LIST
else:
    json_schema = None
   JSON_SCHEMA_VALIDATION_LIST = None
```
5. For the docs, you could put them wherever you want (keep them in main FWs, move them to the new repo, or some of both)

With those changes in place, the code should function just the same and still be integrated with main FireWorks (except that a user will also need to install the ```fw_json_schema``` package). However, it will be clearer that the maintenance of all the JSON files as well as the ``fw_json_schema.py``` is outside of main FireWorks.

ikondov:
Thank you very much for reviewing this pull request. I agree and will do all as you suggest but this will take some more time, probably two or three weeks. I will mark this pull request as "work in progress".

computron:
That is great! Looking forward to it whenever you have the time, thanks as always for your contributions.

computron:
Thanks for making these changes! Merging now.

Issue #399, opened 2020-05-22
error in test "visualize the workflow graph"


ikondov:
Running a workflow check with most recent fireworks 1.9.5 and python-igraph 0.8.2 I get this error message:

```
(python3-venv) ubuntu:~/temp$ lpad check_wflow -i 525 -g combined -f wf.dot
Traceback (most recent call last):
  File "/home/ubuntu/python3-venv/bin/lpad", line 11, in <module>
    load_entry_point('FireWorks==1.9.5', 'console_scripts', 'lpad')()
  File "/home/ubuntu/python3-venv/lib/python3.8/site-packages/fireworks/scripts/lpad_run.py", line 1209, in lpad
    args.func(args)
  File "/home/ubuntu/python3-venv/lib/python3.8/site-packages/fireworks/scripts/lpad_run.py", line 217, in check_wf
    dagf.to_dot(args.dot_file, view=args.view)
  File "/home/ubuntu/python3-venv/lib/python3.8/site-packages/fireworks/utilities/dagflow.py", line 292, in to_dot
    graph = self.copy()
TypeError: __init__() got an unexpected keyword argument '__ptr'
```
I ran the tests but could not detect this error due to other errors (see issue https://github.com/materialsproject/fireworks/issues/398). After fixing issue https://github.com/materialsproject/fireworks/issues/398 one test returns the same error:

```
======================================================================
ERROR: visualize the workflow graph
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 132, in test_dagflow_view
    dagf.to_dot(filename, view='combined')
  File "/home/ubuntu/fireworks/fireworks/utilities/dagflow.py", line 292, in to_dot
    graph = self.copy()
TypeError: __init__() got an unexpected keyword argument '__ptr'
```

Issue #398, opened 2020-05-22
errors during the tests


ikondov:
I have the package python-igraph installed. When I run the tests I get these errors:

```
======================================================================
ERROR: disconnected graph
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 72, in test_dagflow_cut
    DAGFlow.from_fireworks(wfl)
NameError: name 'DAGFlow' is not defined

======================================================================
ERROR: missing input
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 101, in test_dagflow_input
    DAGFlow.from_fireworks(wfl)
NameError: name 'DAGFlow' is not defined

======================================================================
ERROR: wrong links
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 83, in test_dagflow_link
    DAGFlow.from_fireworks(wfl)
NameError: name 'DAGFlow' is not defined

======================================================================
ERROR: loop in graph
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 64, in test_dagflow_loop
    DAGFlow.from_fireworks(wfl)
NameError: name 'DAGFlow' is not defined

======================================================================
ERROR: construct and replicate
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 53, in test_dagflow_ok
    dagf = DAGFlow.from_fireworks(wfl)
NameError: name 'DAGFlow' is not defined

======================================================================
ERROR: clashing inputs
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 119, in test_dagflow_output
    DAGFlow.from_fireworks(wfl)
NameError: name 'DAGFlow' is not defined

======================================================================
ERROR: visualize the workflow graph
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/fireworks/fireworks/utilities/tests/test_dagflow.py", line 128, in test_dagflow_view
[...]

PR #400, opened 2020-05-22
bugfixes: global import of dagflow module; copy() of dagflow replaced

ikondov:
This solves the following issues:
* https://github.com/materialsproject/fireworks/issues/398: missing global import of `DAGFlow` class in the tests
* https://github.com/materialsproject/fireworks/issues/399: caused by some changes in python-igraph `Graph.copy()` method or due to change in python 3.8 deepcopy behavior


ikondov:
I have no clue why the import error is raised. In all three commits my local nosetests either all pass or get skipped depending on whether igraph is installed.

computron:
Merging, although I would have thought that you should be able to import DAGFlow in the setUp

PR #402, opened 2020-07-03
ENH: Filepad accessible via TLS/SSL encrypted connection

jotelha:
Allow TLS/SSL-encrypted connection for FilePad as well (analogous to same mechanism in LaunchPad, code largely copied).


computron:
Thanks!

PR #403, opened 2020-07-03
BUG (?): no dropping of desired input files in diamond dependency

jotelha:
I suggest removing this particular elif clause, as it leads to the diamond dependency issue described in the comment.


computron:
I've been staring at the original ``elif`` clause for a bit to try to figure out it's original purpose, but haven't come up with any reason for it to be there. I do understand the problem you describe. So, I am going to go ahead and accept, thanks!

PR #404, opened 2020-07-03
ENH: Find orphaned Fireworks and remove from db

jotelha:
At some point I encountered the issue of many "orphaned" Fireworks (which did not belong to any workflow any longer) in the database (possibly to some interrupted delete of whole workflows). Many Fireworks commands will refuse to operate on such an inconsistent database. I recommend some maintenance option as suggested here to find (and remove, if desired) such "orphaned" Fireworks, which don't belong to any workflow.

Code and command line interface are largely copied from the `get_fws` and `get_wfs` functions, with according modifications.


fireworks/scripts/lpad_run.py
```diff
@@ -691,6 +692,115 @@ def maintain(args):
...
+        sort = None
+
+    if args.qid:
+        fw_ids = lp.get_fw_ids_from_reservation_id(args.qid)
+        if query:
+            query['fw_id'] = {"$in": fw_ids}
+            fw_ids = lp.get_fw_ids(query, sort, args.max, launches_mode=args.launches_mode)
+    else:
+        fw_ids = lp.get_fw_ids(query, sort, args.max,
+                               launches_mode=args.launches_mode)
```
computron, line 733: This whole section introduces a lot of code duplication from the ```get_fws``` function. Is it possible to create a third (helper) function with this code snippet that both ```get_fws``` and ```orphaned`` can use in order to avoid this duplication? e.g. ```get_fws_helper``` or something

fireworks/scripts/lpad_run.py
```diff
@@ -691,6 +692,115 @@ def maintain(args):
...
+
+    # get_wfs
+    orphaned_fw_ids = []
+    for fw_id in fw_ids:
+        query = {'nodes': fw_id}
+        wf_ids = lp.get_wf_ids(query)
+        if len(wf_ids) == 0:
+            orphaned_fw_ids.append(fw_id)
```
computron, line 741: This seems like it would be really slow for a large FireWorks database, e.g., if the potentially empty query returned 1 million results.

If the query returns more than 100 results or so, it might be quicker to simply make a list of all the ```fw_ids``` present in the database overall (e.g., do an empty query on the ```workflows database```, and get back the nodes parameter). Then you have two set of ```fw_ids``` - the set matching the query, and the set of all fw_ids in the workflows collection. You can use Python's set difference to get the ones that are in the query but not in the workflows collection.

Let me know what you think about this

fireworks/scripts/lpad_run.py
```diff
@@ -691,6 +692,115 @@ def maintain(args):
...
+                    del d['archived_launches']
+                del d['spec']
+            if args.display_format == 'less':
+                if 'launches' in d:
+                    del d['launches']
+            fws.append(d)
+    if len(fws) == 1:
+        fws = orphaned_fw_ids[0]
+
+    if args.remove:  # snippet from lpad.delete_wf
```
computron, line 764: if there is more complex code that is duplicated in this, I think it is better to just have a helper function that both functions reference. Otherwise if we make a fix or change to one of the functions (e.g. lpad.delete_wf) the behavior won't update in this function.

computron:
I have no objections to this function, but think stylistically some improvements should be made to make sure the code stays robust to changes and perhaps to improve the speed. I added some comments to this conversation, let me know what you think.

fireworks/scripts/lpad_run.py
jotelha: Sorry for not touching this for a long time, https://github.com/materialsproject/fireworks/pull/404/commits/72e88889fc09417dd33fcdd5fc719200e3ceac45 introduces such helper functions (split in two, get_fw_ids_helper, and get_fws_helper). I did not test the changes in that commit, relying on the CI here.

fireworks/scripts/lpad_run.py
jotelha: Suggested helper is lp.delete_fws in https://github.com/materialsproject/fireworks/pull/404/commits/72e88889fc09417dd33fcdd5fc719200e3ceac45.

fireworks/scripts/lpad_run.py
jotelha: Very true, some proper joint aggregation query on the fireworks and workflows collections in the db might be  even better, but I won't do that for now.

computron:
Thanks!

PR #405, opened 2020-07-03
ENH: dagflow direct plot

jotelha:
Analogous to Ivan's `to_dot` functionality, this adds a `plot_wf` function to directly generate a graph plot of a workflow. This allows to visually look at the graph directly from some interactive python session (i.e. IPython or Jupyter Notebook) before actually adding the workflow to the LaunchPad, as shown here:

image

To allow for looking at possibly erroneous graphs, I suggest to "weaken" assertions on the dataflow into warnings.

### DIFF
--- a/fireworks/utilities/dagflow.py
+++ b/fireworks/utilities/dagflow.py
@@ -9,10 +9,27 @@ from itertools import combinations
 import igraph
 from igraph import Graph
 
-
 DF_TASKS = ['PyTask', 'CommandLineTask', 'ForeachTask', 'JoinDictTask',
             'JoinListTask']
 
+DEFAULT_IGRAPH_VISUAL_SYTLE = {
+    "bbox": (700, 500),
+    "margin": [200, 100, 200, 100],
+
+    # vertex defaults
+    "vertex_label_angle": -3.14/4.0,
+    "vertex_size": 8,
+    "vertex_shape": 'rectangle',
+    "vertex_label_size": 10,
+    "vertex_label_dist": 4,
+
+    # edge defaults
+    "edge_color": 'black',
+    "edge_width": 1,
+    "edge_arrow_size": 1,
+    "edge_arrow_width": 1,
+    "edge_label_size": 8,
+}
 
 class DAGFlow(Graph):
     """ The purpose of this class is to help construction, validation and
@@ -330,19 +347,20 @@ class DAGFlow(Graph):
 
 
 def plot_wf(wf, view='combined', labels=False, **kwargs):
-    """Plot workflow via igraph.plot.
+    """Plot workflow DAG via igraph.plot.
 
-    Parameters:
-        wf
-        view: str, same as in 'to_dot'
+    Args:
+        wf (Workflow)
+        view (str): same as in 'to_dot'. Default: 'combined'
+        labels (bool): show a FW's name and id as labels in graph
 
     Other **kwargs can be any igraph plotting style keyword, overrides default.
     See https://igraph.org/python/doc/tutorial/tutorial.html for possible
     keywords. See `plot_wf` code for defaults.
 
     Returns:
-        igraph.drawing.Plot"""
-    import matplotlib  # only needed for color-coding with favorite named colors
+        igraph.drawing.Plot
+    """
 
     dagf = DAGFlow.from_fireworks(wf)
     if labels:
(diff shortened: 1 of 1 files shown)


ikondov:
Very nice contribution!

There is a bugfix pull request (https://github.com/materialsproject/fireworks/pull/400) scheduled previously which includes changes in the same class. Maybe you will have to rebase after it is merged.

The reason for the assertions is that I originally designed this class to check workflow's formal correctness (`lpad add -c` or `lpad check_flow`) and to not accept erroneous workflows on the launchpad. But here you have another use case - you have a workflow object and want to plot it without interacting with launchpad, correct? 
The one specific assertion you have converted to a warning
https://github.com/materialsproject/fireworks/pull/405/files#diff-7edc51edcf2ff005dbdd5a7b679feb14L276
https://github.com/materialsproject/fireworks/pull/405/files#diff-7edc51edcf2ff005dbdd5a7b679feb14R290
prevents race conditions for inputs with the same spec keys arriving from multiple parent fireworks. You are right that the case of overriding static input data with dynamic data from FWAction is no race condition. But we need more code to distinguish the two cases and I would be still for an exception in case of a race condition. Another solution is to handle depending on the call context - exception if check is requested and a warning if only visualization. I would appreciate if you could make this, I wanted to do this but had no time ...

The visual style defaults section is a bit too long - probably use the igraph defaults and expose an argument to the user function to to make things more fancy?

computron:
@ikondov can  you clarify a little what these two checks do?

```'The tasks in a workflow step may not share output fields.'```

and:

```'An input field must have exactly one source'```

In particular, what input and output fields refer to. I see some definition for example in ```task_input(task, spec):``` and ```_get_sources``` but those look to be specifically designed for PyTask and ScriptTask and not any custom Firetasks.

Based on just skimming the code (and perhaps some incorrect assumptions), I would probably suggest the following two changes to dagflow:

1. For the checks to really be restricted to ScriptTask and PyTask. This could be done by checking the fw name. The reason I suggest this is that a custom FireTask might just happen to use the field "output". That output field might be an entry to add to a list for the child Firework. Then, it is fine if a Firework has multiple parents with the field "output" - since each parent will add its own unique key to the list. Restricting these checks to ScriptTask and PyTask will make sure you don't get strange failures for custom Fireworks, which may have their own conventions about the use of these fields.

2. To have an option about whether the checks are outputted as warnings or as failures, e.g. ```fail on error```. Then, features like pre-visualizers can use the rest of the code for visualization purposes without having it automatically fail - whereas users doing strict error checking can have the code fail as they desire.

ikondov:
> `'The tasks in a workflow step may not share output fields.'`

This means that two tasks may not write an output to the same spec key. Writing to spec under the same spec key overwrites the output of preceeding Firetasks which is probably not the formally correct behavior. The user should define different names in the firetask parameter `outputs`.

> `'An input field must have exactly one source'`

This means that a Firetask gets a dynamic input, specified in the parameter `inputs`, that is an output of tasks from two or more parent fireworks. This obviously leads to a race condition that must be avoided.

> In particular, what input and output fields refer to. I see some definition for example in `task_input(task, spec):` and `_get_sources` but those look to be specifically designed for PyTask and ScriptTask and not any custom Firetasks.

I thoroughly agree this should be more clearly written. The *input fields* are the list items of parameter `inputs`, and the *output fields* are the list items of parameter `outputs`. I will improve the wording.

> Based on just skimming the code (and perhaps some incorrect assumptions), I would probably suggest the following two changes to dagflow:
> 
>     1. For the checks to really be restricted to ScriptTask and PyTask. This could be done by checking the fw name. The reason I suggest this is that a custom FireTask might just happen to use the field "output". That output field might be an entry to add to a list for the child Firework. Then, it is fine if a Firework has multiple parents with the field "output" - since each parent will add its own unique key to the list. Restricting these checks to ScriptTask and PyTask will make sure you don't get strange failures for custom Fireworks, which may have their own conventions about the use of these fields.

Yes, you are totally right! There should be a check so that the code does not check other Firetasks. But the Firetasks that should be checked are `PyTask` and the dataflow tasks (https://materialsproject.github.io/fireworks/dataflow_tasks.html). The `ScriptTask` and other Firetasks should not be checked. I will do this change.

>     2. To have an option about whether the checks are outputted as warnings or as failures, e.g. `fail_on_error`. Then, features like pre-visualizers can use the rest of the code for visualization purposes without having it automatically fail - whereas users doing strict error checking can have the code fail as they desire.

As I commented previously, visualization and correctness check should be separated. I have written the class mainly to check the correctness. This is why failures of checks should strictly result in exceptions. The export to dot-format (the visualization) was just a 'nice-to-have' function but it was unfortunately misplaced. Visualizing "wrong" workflows is of course quite logical behavior. I will do the refactoring.

BTW: Other things that should be checked but currently not checked are the `_files_in` and `_files_out` keys.

jotelha:
Waiting for https://github.com/materialsproject/fireworks/issues/411 to be resolved, then adapt and rebase.

computron:
Hi @jotelha , I just merged the most recent changes by @ikondov . Let us know if this is what you need!

jotelha:
> The visual style defaults section is a bit too long - probably use the igraph defaults and expose an argument to the user function to to make things more fancy?

Removed most of igraph style defaults and put the remaining quite necessary ones in the header . Should those maybe go to some configuration file instead? The user can specify any custom style parameter via `plot_wf`'s `**kwargs`.

ikondov:
> @ikondov
> 
> > The visual style defaults section is a bit too long - probably use the igraph defaults and expose an argument to the user function to to make things more fancy?
> 
> Removed most of igraph style defaults and put the remaining quite necessary ones in the header . Should those maybe go to some configuration file instead? The user can specify any custom style parameter via `plot_wf`'s `**kwargs`.

I think we need no configuration file. I think it is much more compact now and it can be left in this state.

jotelha:
Sorry for the typo, thanks!

computron:
Thanks!

PR #406, opened 2020-07-03
ENH: stoppable background tasks by simple set_stop_event method on tasks

jotelha:
Tasks wrapped into BackgroundTask might run forever (In my particular case, I use background tasks to establish some ssh connection, i.e. for file transfer to remote machines. The ssh connection stays open as long as Fireworks itself runs, even if the connection-dependent file transfer task has finished, blocking the allocated port indefinitely). Give them an opportunity to exit gracefully by suggesting any such task to implement some `set_stop_event` method and forward the stop event if this method exists. A very simple approach.

### DIFF
--- a/fireworks/core/rocket.py
+++ b/fireworks/core/rocket.py
@@ -84,6 +84,8 @@ def background_task(btask, spec, stop_event, master_thread):
     num_launched = 0
     while not stop_event.is_set() and master_thread.isAlive():
         for task in btask.tasks:
+            if hasattr(task, 'set_stop_event'):
+                task.set_stop_event(stop_event)
             task.run_task(spec)
         if btask.sleep_time > 0:
             stop_event.wait(btask.sleep_time)


computron:
Hi,

The initial (default) purpose of a background task was to run a process repeatedly at some interval - e.g., every 60 seconds or so - while the main task was running. For example, you might want to repeatedly check the progress of an output file every 60  seconds and update a secondary database with the result.

If you are using the BackgroundTask to establish an ssh connection, does it still stay running indefinitely if:

1. You set the ```num_launches``` parameter of the BackgroundTask to 1 (the default of 0 means to run it indefinitely)
2. The code for your background task properly closes the ssh connection upon completion? e.g., uses an explicit ```close()`` command or uses the ```with``` syntax?

 Note that having some kind of external stop event might still make sense (and maybe we want to make this more official by having it be an explicit parameter of BackgroundTask), but just wanted to probe a little more into the situation first.

Issue #401, opened 2020-05-26
lpad webgui problem with gunicorn>=20.0


zhubonan:
`gunicorn.six` is removed in `gunicorn > 20.0`

This line here:

https://github.com/materialsproject/fireworks/blob/6505cd52ca3a7f34b549e6a18c7e2d0d336aa066/fireworks/flask_site/gunicorn.py#L7-L9

still tries to import `iteritem` from it. The easy fix is to tell it import from `six` instead. 

As a result, anyone with `gunicorn>20.0` will get a message saying they need to install `gunicorn`. 

As a general problem, I can see that `setup.py` only requires versions of dependencies to be higher than a particular release. 

https://github.com/materialsproject/fireworks/blob/e33821518acf4dc7276aa161a523a0770a776191/setup.py#L32-L37

If assuming semantic versioning, it would be better to pin them below the next major release number, in case of future non-compatible release breaking things.

PR #408, opened 2020-07-03
BUG: prevents https://github.com/materialsproject/fireworks/issues/401

jotelha:
web gui fails with gunicorn versions >= 20, https://github.com/materialsproject/fireworks/issues/401


computron:
Do you know if the code works if we replace:

```
from gunicorn.six import iteritems
```

with:
```
from six import iteritems
```

The latter would be the preferred solution since we could continue to use the latest gunicorn.

computron:
Thanks!

Issue #411, opened 2020-07-22
Reorganize the DAGFlow class


ikondov:
* Separate the checking functions and the visualizations in `DAGFlow` class
* Remove the `--graph` option in `lpad check_wflow` (or make it strictly alternative to a `--check` option)
* Restrict the dataflow checks to `PyTask` and dataflow Firetasks (https://materialsproject.github.io/fireworks/dataflow_tasks.html)
* Make the messages of checks better understandable 
* Use more specific exceptions instead of assertions so that the caller can catch them if needed

PR #412, opened 2020-07-29
Reorganized DAGFlow class

ikondov:
Solves issue https://github.com/materialsproject/fireworks/issues/411. Only assertions have not been replaced with specific exceptions. Though the issued `AssertionError` can still be trapped by calling functions (see examples in the tests).


computron:
Thanks!

PR #413, opened 2020-08-20
BUG: parameter dest is not required in FileTransferTask

jotelha:
can as well be specified nested below files, i.e. "files": [ {'src': ..., 'dest': ...}]


computron:
Thanks!

PR #414, opened 2020-08-20
ENH: Joint queries on Fireworks and Workflows from command line

jotelha:
This is another loose suggestion for extension: Often, I want to quickly submit a joint query between Fireworks and Workflows, i.e. get an overview on all FIZZLED Fireworks in a particular workflow. Here is an extension to the Launchpad and its command line interface to avoid the "detour" via direct MongoDB queries. One particular consideration for discussion would be how to distinguish between CLI options aimed at selecting Fireworks query and such aimed at selecting Workflows. I chose options prefixed with 'fw' and 'wf' respectively here, i.e.

    lpad get_fws_in_wflows -wfi 38710 -fwq '{"state":"FIZZLED","name":{"$regex":"gmx_mdrun$"}}' -d all

would yield all Fireworks matching the 'fwq' query within the Workflow identified by 'wfi'.

### DIFF
--- a/fireworks/core/launchpad.py
+++ b/fireworks/core/launchpad.py
@@ -770,6 +770,71 @@ class LaunchPad(FWSerializable):
 
         return wf_ids
 
+    def get_fws_ids_in_wfs(self, wf_query=None, fw_query=None, sort=None,
+                           limit=0, count_only=False, launches_mode=False):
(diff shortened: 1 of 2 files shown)


computron:
This will be useful, thanks!

PR #418, opened 2020-10-16
Add support for dataclass Firetasks

utf:
## Summary

Originally submitted in #393, this update fixes some bugs and adds tests.

This pull request extends the Firetasks constructor to support python dataclasses. This allows Firetasks to be explorable when using jupyter notebooks (i.e., tab completion), to support type hinting, and also to have explicit default values. I believe this approach will make documenting and troubleshooting complex Firetasks easier. The changes implemented in this pull request are backwards compatible and should not affect existing Fireworks or their operation. 

### Initialisation

Firetasks can now be initialised using a dataclass approach for specifying required and optional arguments.

```python
@dataclass(repr=False)
class Task(FiretaskBase):
    param: str
    param_a: int = 0

task = Task(param=123)
```

Is equivalent to:
```python
class Task(FiretaskBase):
    required_params = ["param"]
    optional_params = ["param_a"]

task = Task(param=123)
```

In the dataclass case we get type hinting and default arguments for free. 

### Attribute access

Firetask attributes can now be accessed using a dot syntax whether or not they were created with the dataclass syntax. Furthermore, the traditional dictionary style access for attributes also works in both cases. For example, all of these are valid:

```python
task.param = 1
task["param_a"] = 2
task["param"]
task.get("param_a")
task.get("param_a", 123)
```

## Notes

- The `repr=False` syntax is not strictly necessary but means the string representation for both fireworks will be identical.
- I've done more rigorous testing by converting a few key Firetasks in atomate to the dataclass format (specifically `PassCalcLocs`, `VaspToDb`, `WriteVaspFromIOSet`, and `WriteVaspNSCFFromPrev`) and all atomate tests still pass.
- In the previous version of this PR, casting the Firetask to a dictionary directly using `dict(task)` resulted in different results for the dataclass/original approach. This has now been fixed.
- The use of dataclasses requires python 3.6+. Considering fireworks is already running CircleCI only for python 3.7+ I don't think this should be too much of an issue.
- This PR came out of the discussion in this thread hackingmaterials/atomate#377.

## Differences between original and dataclass Firetasks

The behaviour of dataclass Firetasks is different from non-dataclass Firetasks in several minor and subtle ways. 
1. Initialising a fire tasks without the required arguments raises a `TypeError` for dataclasses vs `RuntimeError` for non-dataclasses.
2. Non-dataclass Firetasks can be initialised by passing a dict and no keyword arguments. This behaviour actually allows you to initialise a Firetask with arguments outside of required_params + optional_params. This behaviour is not supported using dataclass Firetasks and instead the `from_dict` method should be used.


mkhorton:
Very happy to see this revived :-)

> The use of dataclasses requires python 3.6+. Considering fireworks is already running CircleCI only for python 3.7+ I don't think this should be too much of an issue.

Regarding this, if it helps, we're only officially supporting 3.6+ in *pymatgen* too. The logic here is (as you say) we're only testing on 3.6+ anyway, and also anyone who is able to run Python 3.x is typically also able to run Python 3.6+, it's very rare to find a user on Python 3 who is not able to upgrade to at least 3.6.

computron:
Note that FireWorks is not intended to be a framework for running pymatgen, and people use it for all sorts of purposes. So the fact that pymatgen is 3.6+ only is not really relevant at all. However, other Python libraries that FireWorks requires (monty in particular) are Python 3+ and moving strongly away from Python 2 is definitely a factor to drop Python 2 support from FireWorks as well.

I am going to put out a message regarding Py2 support on the matsci.org list and see if there is anyone vocalizing strong reasons to keep Py2 support

utf:
It would be relatively easy to modify this PR to make it so that data classes aren't a requirement. That way people could still use fireworks with Py2 just without the dataclass functionality.

computron:
Let's do that for now

mkhorton:
> Note that FireWorks is not intended to be a framework for running pymatgen, and people use it for all sorts of purposes. So the fact that pymatgen is 3.6+ only is not really relevant at all.

My intent with that comment was to demonstrate the thought process for modernizing to 3.6, I'm aware pymatgen is not a dependency of FireWorks.

utf:
Just a note: When I tried to install and test Fireworks in a python 2.7 environment, all the tests failed because the latest monty version is py3 only.

I've therefore updated the requirements so that pip installation works with python2 by setting a max monty version for python versions < 3. I also had to move the dataclass firetasks to a separate file that are only imported if dataclasses are available, as otherwise they cause syntax errors.

Lastly, I added a comparison between dataclass and non-dataclass firetasks in `fireworks/examples/custom_firetasks/dataclasses` and added a pointer to this in the Firetask documentation.

All tests now pass using python 2.7.

utf:
I'm going to close this PR as the main aims were accomplished in https://github.com/materialsproject/jobflow

PR #419, opened 2020-10-27
Fix a bug where the timeout is not honored in `rlaunch multi`

zhubonan:
There was a bug where the timeout is not honoured in multi-launch.

The problem is because the new rapid-fire process will be launched if there are other processes running, despite there are no active or future jobs in the FireServer. However, the new rapid-fire should use an updated timeout based on elapsed time, rather than the original timeout value.

To avoid very short launches, I also made it such that the process will stand down if there is less than 3% of the original timeout available.


zhubonan:
@utf Hi Alex, could you please take a look at this PR? Thanks!

utf:
Looks good to me. Maybe @computron can confirm.

PR #421, opened 2020-10-27
Don't fork a process to start the web server -- Fix #416

### LINKED ISSUE #416: default multiprocessing change
As of now running the following command: "lpad -l my_launchpad.yaml webgui" will result in the following error:
File "/Users/me/miniconda3/bin/lpad", line 11, in <module>
    load_entry_point('FireWorks', 'console_scripts', 'lpad')()
  File "/Users/me/repos/fireworks/fireworks/scripts/lpad_run.py", line 1434, in lpad
    args.func(args)
  File "/Users/me/repos/fireworks/fireworks/scripts/lpad_run.py", line 686, in webgui
    p1.start()
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/context.py", line 283, in _Popen
    return Popen(process_obj)
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/Users/me/miniconda3/lib/python3.8/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: cannot pickle '_thread.lock' object

This beavior is observed under python >= 3.8.X  and a macOS Catalina environment. I suspect that the recent feature changes in python 3.8, more precisely, on MacOS, the spawn start method is not used by default for multiprocessing. I believe a straightforward fix would be to explicitly define the start method as 'fork' using: 
from multiprocessing import set_start_method
set_start_method('fork')

This has fixed the problem. However, note that the following UserWarning will appear: 
/Users/me/miniconda3/lib/python3.8/site-packages/pymongo-3.11.0-py3.8-macosx-10.9-x86_64.egg/pymongo/topology.py:161: UserWarning: MongoClient opened before fork. Create MongoClient only after forking. See PyMongo's documentation for details: https://pymongo.readthedocs.io/en/stable/faq.html#is-pymongo-fork-safe

I believe this is coming from  "app.lp = get_lp(args)" Line 662 of the lpad_run.py. On unix machines, the multiprocessing spawns processes using fork, where instances of MongoClient should not be copied from a parent process to a child process. While, as defined in lpad_run.py the MOngoClient instance is not created by the process, and the global MongoClient is being copied by the process instead.

1fish2:
Fix the `TypeError: cannot pickle '_thread.lock' object` Issue #416 by not spawning a webserver process.

It could just call `webbrowser.open()` without additional parallelization, but this spawns a thread to call it to preserve the delay that gives the web server more time to get ready to serve pages.

I tested this on a Mac with Chrome, both in the cases where Chrome was already running and not. I didn't test other OSs or browsers, but it's using stock Python APIs. I tested this in Python 3.8.6. It should be fine in other Python releases, although the Python docs listed the `threading` module under "Optional Operating System Services" in Python 2.7 -- it's certainly present in the Mac Python 2.7.16.

Note: When `webbrowser.open(url)` needs to launch Chrome, Chrome opens the Fireworks `url` in a new window then restores its previous tabs in another window right on top of the Fireworks window, kind of hiding it.

(This branch forked from the `v1.9.6` tag since I don't know the state of the unreleased changes in master.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,6 @@ fw_tutorials/firetask/addition_task2.py
 fw_tutorials/firetask/words.txt
 
 my_launchpad.yaml
+.python-version
+FW.json
 .DS_Store
--- a/fireworks/scripts/lpad_run.py
+++ b/fireworks/scripts/lpad_run.py
@@ -535,6 +535,13 @@ def set_priority(args):
     lp.m_logger.info("Finished setting priorities of {} FWs".format(len(fw_ids)))
 
 
+def _open_webbrowser(url):
+    """Open a web browser after a delay to give the web server more startup time."""
+    import webbrowser
+    time.sleep(2)
+    webbrowser.open(url)
+
+
 def webgui(args):
     from fireworks.flask_site.app import app
     app.lp = get_lp(args)
@@ -554,14 +561,11 @@ def webgui(args):
             app.BASE_Q_WF["state"] = app.BASE_Q["state"]
 
     if not args.server_mode:
-        from multiprocessing import Process
-        p1 = Process(
-            target=app.run,
-            kwargs={"host": args.host, "port": args.port, "debug": args.debug})
+        from threading import Thread
+        url = "http://{}:{}".format(args.host, args.port)
+        p1 = Thread(target=_open_webbrowser, args=(url,))
         p1.start()
-        import webbrowser
-        time.sleep(2)
-        webbrowser.open("http://{}:{}".format(args.host, args.port))
+        app.run(host=args.host, port=args.port, debug=args.debug)
         p1.join()
     else:
         try:


computron:
Thanks!

PR #423, opened 2020-11-18
Functionality to optionally checkpoint Fireworks in response to signals

helgammal:
_This is work which was coordinated with @mkhorton and @shreddd._

## Changes:
- Adding new `checkpoint` abstract method to `FiretaskBase`.
- Adding new `CHECKPOINTED` status to `Firework.STATE_RANKS`. 
- Implementing a new `checkpoint_fw` to `launchpad` which looks for `RUNNING` Fireworks that match a given ID and changes their status to `CHECKPOINTED`.
- Implementing `checkpoint` in `ScriptTask` which depends on a `Launchpad` object being passed to the ScriptTask (this can currently be done by passing `_add_launchpad_and_fw_id` through the YAML config).

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -209,8 +209,8 @@ class Firework(FWSerializable):
     A Firework is a workflow step and might be contain several Firetasks.
     """
     STATE_RANKS = {'ARCHIVED': -2, 'FIZZLED': -1, 'DEFUSED': 0, 'PAUSED': 0,
-                   'WAITING': 1, 'READY': 2, 'RESERVED': 3, 'RUNNING': 4,
-                   'COMPLETED': 5, 'CHECKPOINTED': 6}
+                   'WAITING': 1, 'READY': 2, 'RESERVED': 3, 'RUNNING': 4, 
+                   'CHECKPOINTED': 5, 'COMPLETED': 6}
 
     # note: if you modify this signature, you must also modify LazyFirework
     def __init__(self, tasks, spec=None, name=None, launches=None, archived_launches=None,
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -73,6 +73,10 @@ class FiretaskBase(defaultdict, FWSerializable):
                             self.__class__, k, allowed_params))
 
     @abc.abstractmethod
+    def checkpoint(self, signum, stack):
+        raise NotImplementedError("You must have a checkpoint implemented!")
+
+    @abc.abstractmethod
     def run_task(self, fw_spec):
         """
         This method gets called when the Firetask is run. It can take in a
@@ -204,10 +208,9 @@ class Firework(FWSerializable):
     """
     A Firework is a workflow step and might be contain several Firetasks.
     """
-
     STATE_RANKS = {'ARCHIVED': -2, 'FIZZLED': -1, 'DEFUSED': 0, 'PAUSED': 0,
                    'WAITING': 1, 'READY': 2, 'RESERVED': 3, 'RUNNING': 4,
-                   'COMPLETED': 5}
+                   'COMPLETED': 5, 'CHECKPOINTED': 6}
 
     # note: if you modify this signature, you must also modify LazyFirework
     def __init__(self, tasks, spec=None, name=None, launches=None, archived_launches=None,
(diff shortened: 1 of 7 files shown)


mkhorton:
Hi @computron, I've met with @helgammal to go over this PR. I think the functionality is great, implemented in a clean way and is ready for your review.

There are two points I think should be addressed before a potential merge:

* There needs to be a brief paragraph on this feature added to the documentation. Specifically to talk about the new `checkpoint` abstract method, highlighting `ScriptTask` as the example, the role that the user signal plays, and also mentioning how any subprocess spawned by FireWorks has to use `preexec_fn=ignore_SIGUSR1` so that the signal is correctly handled by FireWorks and not the subprocess since this is a subtle point that might be missed otherwise.

* Currently, for users who want to enable check-pointing, two extra lines have to be added to the relevant `run_task`. We've discussed alternatives to get these lines out of `run_task` so that the burden on the user is reduced, so that `run_task` does not need to be modified.

helgammal:
Thanks for the write-up @mkhorton! I've taken out the signal setting into `rocket.py` from `script_task.py`; seems to work as expected. 

To document, I added a write-up to indicate the points you mentioned (as an appendix entitled "Appendix 3: checkpointing within Firetasks" [here](https://github.com/helgammal/fireworks/blob/checkpointing_fireworks/docs_rst/guide_to_writing_firetasks.rst#appendix-3-checkpointing-within-firetasks)).

fireworks/core/rocket.py
```diff
@@ -259,6 +260,8 @@ def run(self, pdb_on_exception=False):
                     t.fworker = self.fworker
 
                 try:
+                    if getattr(t, 'checkpoint', None) and callable(t.checkpoint):
```
mkhorton, line 263: Can remove `None`

fireworks/core/firework.py
```diff
@@ -204,10 +208,9 @@ class Firework(FWSerializable):
     """
     A Firework is a workflow step and might be contain several Firetasks.
     """
-
     STATE_RANKS = {'ARCHIVED': -2, 'FIZZLED': -1, 'DEFUSED': 0, 'PAUSED': 0,
                    'WAITING': 1, 'READY': 2, 'RESERVED': 3, 'RUNNING': 4,
-                   'COMPLETED': 5}
+                   'COMPLETED': 5, 'CHECKPOINTED': 6}
```
mkhorton, line 213: @computron comment: rank is used to set priority, so a checkpointed state should be less than completed

fireworks/core/launchpad.py
```diff
@@ -1082,6 +1082,26 @@ def resume_fw(self, fw_id):
...
 
+    def checkpoint_fw(self, fw_id):
+        """
+        Given the firework id, mark firework as checkpointed and refresh the workflow
+
+        Args:
+            fw_id(int): firework id
+        """
+        allowed_states = ['RUNNING']
+        f = self.fireworks.find_one_and_update(
```
mkhorton, line 1093: @computron had some discussion here that maybe the state should be updated in the `launches`

utf:
Hi @mkhorton and @helgammal, could you please give an update on the status of this PR?

helgammal:
@utf : There wasn't a ton of updates since I last left this off in December. Before I started on other projects in January, @computron and I had corresponded over email about this PR. His main note was to modify the handling of how the representative launch is inferred but maybe he can confirm a little more about that over this PR. Here was the last email he'd sent to me and @mkhorton: 
```
For representative launch I think CHECKPOINTED is good. But I think RUNNING should be higher priority. i.e., if there is a launch that is currently RUNNING, then the Firework should be listed as RUNNING (and consequently, the workflow is also RUNNING). In that case, the checkpointed launch (if one exists) takes a back seat until the alternate launch is no longer running.

For allows for pulling to make sure that a Firework that is in CHECKPOINTED state can be pulled via:
fireworks.core.launchpad.LaunchPad._get_a_fw_to_run

(not just READY fws).
```

mkhorton:
Hi both, basically we have to pick this up again to finish up a few loose ends (namely, _resuming_ from a checkpoint FireWork, and testing integration, e.g. with VASP, with the FireWorks dashboard, etc.)

If there was opportunity or interest from @hamdy to do more work on this, I think that would be great, but if you've moved onto other things I can try and pick this up myself.

computron:
Closing this PR for now as it is stale and no longer eligible for merge; feel free to re-open if this functionality is planned for the future.

PR #428, opened 2020-12-26
Use is_alive in favour of isAlive for Python 3.9 compatibility.

### LINKED ISSUE #425: Python 3.9 fails. isAlive() renamed to is_alive()
It seems that Python 3.9 has renamed isAlive to is_alive. This affects directly to core/rocket.py

[https://bugs.python.org/issue37804](https://bugs.python.org/issue37804)

Thanks

tirkarthi:
Fixes #425


computron:
Thank you!

PR #432, opened 2021-02-25
add gpu option to slurm template

lastephey:
Hi FireWorks developers,

This PR adds a GPU option to the SLURM template

I'm at NERSC and thinking about running FireWorks on Perlmutter. I wanted to make it easier for users to do this without manually editing the SLURM template themselves.

Best regards,
Laurie


computron:
Thank you!

PR #435, opened 2021-04-02
Updating urls in the templates

fekad:
To allow to use url prefixes all url should contain the base_url path as prefix. Flask has the `url_for` helper function that can be used not only for the static files but for all urls.


fekad:
Hi @computron, @utf! Could you please have a look at this pull request? We would like to use the GUI of the fireworks inside a jupytehub/jupyterlab framework (https://github.com/fekad/jupyter-fireworks-proxy) which doesn't work with hardcoded absolute paths.

fireworks/flask_site/app.py
```diff
@@ -17,12 +17,11 @@
 from fireworks.flask_site.util import jsonify
 
 app = Flask(__name__)
-app.use_reloader = True
+# app.use_reloader = True
```
utf, line 20: What is the impact of disabled the reloader?

fireworks/flask_site/app.py
```diff
@@ -407,4 +406,16 @@ def simple(coll, interval, num_intervals):
 
 
 if __name__ == "__main__":
-    app.run(debug=True, port=8080)
+    # https://dlukes.github.io/flask-wsgi-url-prefix.html
+
+    from fireworks import LaunchPad
+    from werkzeug.middleware.dispatcher import DispatcherMiddleware
```
utf, line 412: Is this a new dependency or something that is already required by Flask?

fireworks/flask_site/app.py
```diff
@@ -407,4 +406,16 @@ def simple(coll, interval, num_intervals):
...
+    # from werkzeug.wrappers import Response
```
utf, line 413: Remove unused code.

utf (commented):
Hi @fekad, the aims of this PR seem great. I'm not very familiar with the fireworks flask site, so I will have to defer to @computron for this specifics of this PR.

In general, I think it looks good but I have some questions which I added as a code review.

fireworks/flask_site/app.py
fekad: As far as I remember the `use_reloader` is useful for debugging but by default, you would avoid it. This option is for continuously monitor the code of the module and restart the server when there is a change.  Usually, if the debug flag is set the server will automatically reload for code changes:
```python
app.run(debug=True, port=8080)
```

fireworks/flask_site/app.py
fekad: Flask has been built on top of werkzeug so it is already required.

fireworks/flask_site/app.py
fekad: Generally, I would agree with you but I left there to avoid any confusion. The Response class of Flask used before is a subclass of the Response class of werkzeug. So technically it is ok to use but not necessary to use the flask's version at all and this is why it is in a comment. 
In conclusion, this whole part is just an example of using a non-root URL path. Because it is a "special" case probably it is not even the best place for it :) and just keep:
```python
    from fireworks import LaunchPad

    app.lp = LaunchPad()
    app.run(debug=True, port=8080, threaded=False)
```
or the whole thing could just go into the comment:
```python
    from fireworks import LaunchPad
    
    # https://dlukes.github.io/flask-wsgi-url-prefix.html
    # from werkzeug.middleware.dispatcher import DispatcherMiddleware
    # from werkzeug.wrappers import Response
    #
    # app.wsgi_app = DispatcherMiddleware(
    #     Response('Not Found', status=404),
    #     {'/my-app': app.wsgi_app}
    # )

    app.lp = LaunchPad()
    app.run(debug=True, port=8080, threaded=False)
```

utf: Ok, thanks for the information. I'm in favour of your second example. Could you please comment out the whole block and add an additional comment say that this is an example of using a non-robot URL path.

fireworks/flask_site/app.py
utf: Ok, I will defer to @computron on whether this is safe to comment out. But if it is then I would remove the entire line rather than commenting it.

davidwaroquiers:
Thanks @fekad for pushing this, @utf and @computron we are actually giving a tutorial on fireworks in the framework of Nomad next week (11 May) and it would be nice to have this pull request merged by that time. I don't know if this is feasible.

Best,

David

fireworks/flask_site/app.py
fekad:
```suggestion

```

fireworks/flask_site/app.py
fekad: @utf Thanks for the feedback! I modified the PR according to your suggestions.

fekad:
Hi @computron, do you have any thoughts on this one?

fekad:
Hi @computron, do you have any thoughts on this one? If you are simply not interested in please let me know and I close this PR and depreciate [jupyter-fireworks-proxy](https://github.com/modl-uclouvain/jupyter-fireworks-proxy) package from pip.

fekad:
Closing due to the lack of interest!

PR #437, opened 2021-04-21
add mpworkshop's wf_to_graph + SSOT for state_to_color

janosh:
This PR adds `wf_to_graph` [from here](https://git.io/JO6L8) to `fireworks/utilities/dagflow.py` as discussed in [atomate#638](https://github.com/hackingmaterials/atomate/issues/638#issuecomment-824273303).

It also refactors `state_to_color` to have a single source of truth in `fireworks/features/fw_report.py`. Let me know if this is unwanted or should be placed somewhere else.


utf:
Great. Thank you!

PR #445, opened 2021-05-08
Use current_thread instead of currentThread method that was deprecated in Python 3.10.

tirkarthi:
Ref : python/cpython#25174


utf:
Thanks!

tirkarthi:
Thanks @utf

PR #461, opened 2021-10-26
Some format string fixes and upgrades

janosh:
Only the first two commits in this PR are actual fixes.

The other two are just
- upgrading the remaining `"".format()` strings to f-strings (which make errors like that fixed in 2nd commit impossible) (not sure why `pyupgrade` left some of them) and
- joining awkwardly split strings (e.g. ``"Gunicorn is required for server mode. " "Install using `pip install gunicorn`."`` -> ``"Gunicorn is required for server mode. Install using `pip install gunicorn`."``).


computron:
Thanks!

PR #464, opened 2021-11-15
Implement __len__ + __iter__ for Firework + Workflow

janosh:
Also adds a few type hints.

Added a test for the `Workflow` `__len__` and `__iter__ ` but couldn't find the test class for `Fireworks`.

### DIFF
--- a/fireworks/core/rocket.py
+++ b/fireworks/core/rocket.py
@@ -1,4 +1,8 @@
+from typing import Dict, Union
+
 from monty.os.path import zpath
+
+from fireworks.core.fworker import FWorker
 
 """
 A Rocket fetches a Firework from the database, runs the sequence of Firetasks inside, and then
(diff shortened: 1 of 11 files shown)


janosh:
Is the `launch_id` in `fireworks/core/rocket.py` `int` or `str`?

fireworks/core/rocket.py
```diff
@@ -42,32 +46,30 @@
 __date__ = "Feb 7, 2013"
 
 
-def do_ping(launchpad, launch_id):
+def do_ping(launchpad: LaunchPad, launch_id: int) -> None:
```
janosh, line 49: Is this `launch_id` `int` or `str`? Went with `int` but wasn't sure.

computron: launch_id is an int, thanks!

PR #472, opened 2022-01-05
use blueprint for lpad gui

tschaume:
Converting lauchpad GUI to a Flask Blueprint to allow for it to be mounted at a different prefix/application root @mkhorton @shreddd


mkhorton:
Thanks @tschaume. Can I confirm that, to your knowledge, there are no breaking changes in this PR?

@computron This PR is to allow us to more easily deploy FW dashboards for multiple people on NERSC infrastructure but should otherwise not affect dashboards for regular users.

tschaume:
@mkhorton There shouldn't be breaking changes but the commented-out part in `fireworks/scripts/lpad_run.py` needs to be tested.

computron:
@mkhorton @tschaume I'm trying to catch up on some PRs - could you let me know the current status of this? It looks like some tests are failing but maybe they just need a quick rerun or check

tschaume:
@shreddd and @mkhorton have used the blueprint successfully already I think. Based on what's in this PR, [this part](https://github.com/materialsproject/fireworks/pull/472/files#diff-9284194549e03edb7350e332bbd91f10734e1990ed7a90b6e98558e4d8cbd62a) needs reimplementation to work with blueprints.

PR #480, opened 2022-02-03
Better CLI errors on missing config files

janosh:
At the moment, when not setting `FW_config.yaml` or giving a non-existant path to a config file like `-l path/to/nowhere/my_launchpad.yaml`, the CLIs print a rather unhelpful error message:

```py
  File "/rds/user/jr769/hpc-work/.venv/py39/bin/qlaunch", line 8, in <module>
    sys.exit(qlaunch())
  File "/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/site-packages/fireworks/scripts/qlaunch_run.py", line 326, in qlaunch
    do_launch(args)
  File "/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/site-packages/fireworks/scripts/qlaunch_run.py", line 76, in do_launch
    queueadapter = load_object_from_file(args.queueadapter_file)
  File "/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/site-packages/fireworks/utilities/fw_serializers.py", line 385, in load_object_from_file
    f_format = filename.split(".")[-1]
AttributeError: 'NoneType' object has no attribute 'split'
```

This PR changes the behavior to

```py
FileNotFoundError: launchpad_file 'path/to/nowhere/my_launchpad.yaml' does not exist!
```

I also added tests for all 4 CLIs to ensure this behavior doesn't break in the future.


janosh:
Still needs tests for the new `lpad` config file validation and manual testing in a real world setting.

janosh:
Alright, we're getting there. I think these changes will make the CLIs more user friendly when not passing in required config files or providing non-existing file paths or trying to connect to a database on a host:port where none is running.

@arosen93 I've done some local testing and everything seems to work. But after the issues in #476 and considering I made quite a few changes, I'd like to be more thorough with testing this time. My own usage of FW might be quite limited compared to yours. If you have a few minutes, would you be able to install these changes and see if you encounter any problems running your usual commands?

```sh
pip uninstall fireworks -y
pip install https://github.com/janosh/fireworks/zipball/better-cli-errors-on-missing-config-files
```

fireworks/fw_config.py
```diff
@@ -17,7 +17,7 @@
 NEGATIVE_FWID_CTR = 0
 
 # this is where load_object() looks for serialized objects
-USER_PACKAGES = ["fireworks.user_objects", "fireworks.utilities.tests", "fw_tutorials", "fireworks.features"]
+USER_PACKAGES = ["fireworks.user_objects", "fw_tutorials", "fireworks.features"]
```
janosh, line 20: @computron Is this change ok to do? I'm not sure of its implications.

My reason for doing this is:

I added some new tests that use `pytest` rather than `unittest` which now causes errors:

```py
lpad found /Users/janosh/Repos/dielectric-frontier/FW_config.yaml
/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/site-packages/fireworks/utilities/fw_serializers.py:361: UserWarning: None in fireworks.utilities.tests.test_visualize cannot be loaded because of No module named 'pytest'. Skipping..
  warnings.warn(f"{m_object} in {mod_name} cannot be loaded because of {str(ex)}. Skipping..")
  File "/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/site-packages/fireworks/utilities/fw_serializers.py", line 354, in load_object
    m_module = importlib.import_module(mod_name)
  File "/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/rds/user/jr769/hpc-work/.venv/py39/lib/python3.9/site-packages/fireworks/utilities/tests/test_visualize.py", line 1, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

The errors are triggered by this `load_function()`

https://github.com/materialsproject/fireworks/blob/f9f3fe73d2de6ee724ca6e14832ab6b2b04a66ca/fireworks/utilities/fw_serializers.py#L350-L362

Trying to import pretty much every package imported anywhere in Fireworks, even in test files.

Why is that necessary to load a simple YAML file? This seems like a bad idea in general. Second, why include test files here at all? Wouldn't the regular package files be enough to check for imports?

arosen93:
@janosh, sure I'd be happy to. Will aim to do so at the end of the day, CA time. From our prior conversation, it also sounds like if `FW_CONFIG_DIR` is wrong, the improved error message should be present? That's a scenario where I can see it being particularly helpful, so I'll confirm.

janosh:
Thanks!

> From our prior conversation, it also sounds like if `FW_CONFIG_DIR` is wrong, the improved error message should be present?

Yes.

arosen93:
Things seem to work okay for me based on a few common test cases. I will admit, in practice I have never used `-l`, although I do swap config directories all the time with `-c`. That's probably why I never (accidentally) ran into the "unhelpful" CLI error.

janosh:
@arosen93 Thanks for checking! I prefer to use the `-c` flag as well. But then I'm curious, if you've ever misspecified the path to your config directory, you should have gotten this other not very helpful error message which annoyingly only appears after 30 seconds of waiting, right?

```py
pymongo.errors.ServerSelectionTimeoutError: localhost:27017: [Errno 61] Connection refused, Timeout: 30s, Topology Description: <TopologyDescription id: 62022707e3e544f54d361c5d, topology_type: Unknown, servers: [<ServerDescription ('localhost', 27017) server_type: Unknown, rtt: None, error=AutoReconnect('localhost:27017: [Errno 61] Connection refused')>]>
```

With this change the error appears after half a second and says

```py
ValueError: FireWorks was not able to connect to MongoDB at localhost:27017. Is the server running? The database file specified was None. Type "lpad init" if you would like to set up a file that specifies location and credentials of your Mongo database (otherwise use default localhost configuration).
```

I cut the wait time to half a second since Fireworks, if it can't find a launchpad file in the `-c`, defaults to creating a `Launchpad` with default parameters, i.e. one that tries to connect to a local Mongo DB in which case the connection should either be very quick to establish or not possible at all if none is running.

Incidentally, this discussion caused me to find a small error in the code:

```diff
- os.path.exists(os.path.join(args.config_dir, f"my_{file_name}.yaml"))
+ os.path.exists(os.path.join(args.config_dir, f"my_{config_file}.yaml"))
```

@computron If the [change in `fireworks/fw_config.py` I highlighted above](https://github.com/materialsproject/fireworks/pull/480/files/18feac8244e69d89e9b4a1696776ecb2ee38c7c7..5429d9a920b5300180a733569679bb9a5a174131#diff-4a8d4727a1f33694bc679df97d84314eb330e01ed819c5c58acadc1a12e6c7a0) is acceptable, I think this is ready for review.

PR #483, opened 2022-03-03
qlaunch not working on a fresh install

jmmshn:
# qlaunch not working on a fresh install

It appears a `pytest` import was added in the path of `USER_PACKAGES`

https://github.com/materialsproject/fireworks/blob/51aa296dff9cd0411ad9165973dd4bc5c4a5c7aa/fireworks/fw_config.py#L20

And the part of the code that is meant to deal with the exception has a call to `traceback.print_exc(ex)`
which is treating the `ex` exception as the `limit` parameter. 

https://github.com/materialsproject/fireworks/blob/635fca928777610e5933a4f197c0be024aaa82ef/fireworks/utilities/fw_serializers.py#L362

Code for `print_exc`:
```
Signature: traceback.print_exc(limit=None, file=None, chain=True)
Source:   
def print_exc(limit=None, file=None, chain=True):
    """Shorthand for 'print_exception(*sys.exc_info(), limit, file)'."""
    print_exception(*sys.exc_info(), limit=limit, file=file, chain=chain)
File:      ~/miniconda3/envs/mp/lib/python3.9/traceback.py
Type:      function
```

Full output of the error
```
/g/g20/shen9/repos/mp/fireworks/fireworks/utilities/fw_serializers.py:361: UserWarning: None in fireworks.utilities.tests.test_visualize cannot be loaded because of No module named 'pytest'. Skipping..
  warnings.warn(f"{m_object} in {mod_name} cannot be loaded because of {str(ex)}. Skipping..")
Traceback (most recent call last):
  File "/g/g20/shen9/repos/mp/fireworks/fireworks/utilities/fw_serializers.py", line 354, in load_object
    m_module = importlib.import_module(mod_name)
  File "/g/g20/shen9/.conda/envs/mp2/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/g/g20/shen9/repos/mp/fireworks/fireworks/utilities/tests/test_visualize.py", line 1, in <module>
[...]

PR #486, opened 2022-04-19
Update requirements.txt

hadeelsharaf:
to fix an issue when running the workflows on windows.

Error from Traceback : 
```
  File "D:\PY3\lib\site-packages\fireworks\utilities\tests\test_visualize.py", line 1, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ flask==1.1.1
 flask-paginate==0.5.5
 gunicorn==20.0.4
 tqdm==4.41.0
+pytest==7.1.1


arosen93:
We probably don't want to add pytest as a requirement and should fix the source of the issue instead. @janosh, this seems related to a change you recently made, right?

janosh:
@arosen93 Yes, this looks like the same error discussed in https://github.com/materialsproject/fireworks/pull/483#issuecomment-1059212580. I believe the error is already fixed on `main` so the first thing to try would be to release a new version. Not sure if anyone besides @computron can do that? @jmmshn [also pinged](https://github.com/materialsproject/fireworks/pull/483#issuecomment-1090390858) @mkhorton, so maybe he can too?

Regarding this PR, `setuptools` doesn't actually look at `requirements.txt` so adding `pytest` there wouldn't help. It would need to be in `setup.py`.

arosen93:
Thanks for the reply. I will chat with folks here to see how we can get a new version released. This PR can probably be closed thereafter.

ardunn:
> Thanks for the reply. I will chat with folks here to see how we can get a new version released. This PR can probably be closed thereafter.

I believe @computron or @mkhorton can make a release, tagging them here in case one of them is interested

janosh:
Closing this as already fixed.

PR #494, opened 2023-02-04
Fix using URI mode to provide DB info

mjwen:
Previous code and docs for using the URI mode of a DB is based on mLab, which has been acquired by MongoDB Atlas. This PR fixes the URI mode. 

- now, when using the URI mode, the database name need to be explicitly provided 
- updated the corresponding docs


janosh:
Thanks @mjwen! 👍

PR #498, opened 2023-04-11
add pbs filesystems option for alcf polaris

lastephey:
Dear FireWorks developers,

In testing FireWorks on ALCF Polaris I found that they require the PBS filesystems flag. I have tested it in my own local adapter and now am adding it to your PBS template.

Please let me know if you have any questions. 

Thank you,
Laurie


janosh:
Thanks @lastephey!

PR #499, opened 2023-04-11
Enable launchpad and fireworker config be passed via json string

JBlaschke:
This is a relatively simple addition to the `rlaunch` CLI tool: When `--json` is provided, then the inputs to `-l` and `-w` are JSON-formatted strings. This is mainly to allow for `LaunchPad` and `FWorker` to be specified in a python script, without having to generate these files first.

An example workflow can be found here: https://github.com/JBlaschke/alcc-recipes/blob/jpb/fireworks/fireworks/test_3_python/fw.py#L26 where a python script manages a 2-node and a 4-node slurm job. In this case `fireworker_1`, and `fireworker_2` need to be passed to the slurm script in addition to `launchpad`. With the addition of a `--json` flag, this can be handled from the command line without needing to manage additional files.

Tagging @lastephey and @computron

### DIFF
--- a/fireworks/scripts/rlaunch_run.py
+++ b/fireworks/scripts/rlaunch_run.py
@@ -117,6 +117,8 @@ def rlaunch(argv: Optional[Sequence[str]] = None) -> int:
     parser.add_argument("--loglvl", help="level to print log messages", default="INFO")
     parser.add_argument("-s", "--silencer", help="shortcut to mute log messages", action="store_true")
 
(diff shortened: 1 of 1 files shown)


computron:
Thanks! But I see a typo:

launchpad_generator = lambda x: LaunchPad.form_file(x)

Can you fix and also run through the confirm the previous options are working correctly after the changes?

JBlaschke:
@computron thanks for catching that! Fixed.

janosh:
@JBlaschke Could you add a test for this?

PR #503, opened 2023-05-21
[BUG FIX] remove `_id` from `archived_launches`

jmmshn:
I noticed some restarted jobs not rendering because the `_id` fields are pulled from the `archived_launches` which results in the following error in the `webgui`:

```
  File "/Users/shen9/local/Caskroom/miniconda/base/envs/mp/lib/python3.9/site-packages/flask/json/provider.py", line 122, in _default
    raise TypeError(f"Object of type {type(o).__name__} is not JSON serializable")
```

And does not allow the page to be opened. This PR should fix that problem.

### DIFF
--- a/fireworks/flask_site/app.py
+++ b/fireworks/flask_site/app.py
@@ -161,7 +161,7 @@ def get_fw_details(fw_id):
     # to control their display
     fw = app.lp.get_fw_dict_by_id(fw_id)
     for launch in fw["launches"]:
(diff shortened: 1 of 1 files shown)


janosh:
Maybe casting to `str` would be better?

jmmshn:
The `_id` is never used for any of our codes.  But I'm OK with casting it to `str` either way.

PR #504, opened 2023-07-20
change 'python-igraph' dependency to 'igraph'

szhorvat:
Please see https://github.com/igraph/python-igraph/issues/699 for the explanation.


janosh:
Thanks!

PR #505, opened 2023-08-03
Fix `webgui` KeyError

jmmshn:
# Two minor bug fixes

- The WebGUI crashes when some configuration variables cannot be access,  the `.get(..., None)` pattern fixes this and the workflows render for me.
- The `isort` version in the current pre-commit crashes but bumping it to `5.11.5` fixes this.


jmmshn:
@JBlaschke do you have merge access to Fireworks, can you havea look at this?

PR #511, opened 2023-09-04
Remove Unused Dependency: Six

gdrosos:
## Summary

Hello @computron, @janosh,

I hope you're all doing well! I've just opened this pull request that proposes the removal of the unused `six` dependency from the `setup.py` configuration file. It's part of an ongoing research endeavor focusing on the identification and elimination of code bloat within software projects. Your insights on this would be really valuable.

## Rationale

The `six` library was initially introduced to the project in commit 3054682 for its use in` fireworks/core/firework.py`. However, it was subsequently removed from the source code in commit a0db49d and also removed from the setup configuration file in pull request #464. Interestingly, it was reintroduced in commit 26786f5 in the `setup.py` without being directly invoked in the source code.

Given the above, removing this unused dependency will significantly benefit the project. It will not only reduce the overall size of the installed packages but also simplify the dependency tree and mitigate potential security risks.

## Changes

- Removed the `six` dependency from `setup.py`.

## Impact

- **Reduced Package Size**: The removal of this unused dependency will lead to a decrease in the overall size of the installed packages.
- **Simplified Dependency Tree**: Fewer dependencies make the project easier to maintain and can speed up installation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,6 @@ if __name__ == "__main__":
             "ruamel.yaml>=0.15.35",
             "pymongo>=3.3.0",
             "Jinja2>=2.8.0",
-            "six>=1.10.0",
             "monty>=1.0.1",
             "python-dateutil>=2.5.3",
             "tabulate>=0.7.5",


janosh:
Thanks @gdrosos !

gdrosos:
Thank you very much @janosh for the immediate response! Have a nice day!

PR #551, opened 2025-09-30
`FIZZLE` Fireworks whose tasks can't deserialize and move on in the queue

janosh:
prev would try to checkout failing FW endlessly, stalling the queue

- change `LaunchPad._get_a_fw_to_run` class to handle deserialization failures more robustly, marking fireworks as FIZZLED when deserialization fails
- add tests for deserialization failures in `test_deserialization.py` confirming un-deserializable fireworks are FIZZLED and the queue continues

this issue was first spotted and the behavior change proposed by @mkhorton

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -15,7 +15,7 @@ import pprint
 from collections import defaultdict
 from copy import deepcopy
 from datetime import datetime
-from typing import Any, Iterator, NoReturn, Sequence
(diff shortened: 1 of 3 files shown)


janosh:
good point! added a new test to verify this in [8ce3aa](https://github.com/materialsproject/fireworks/pull/551/commits/18ce3aa25d9dfec23f019b22e2b215633f4dcf07)

PR #554, opened 2025-10-24
Speedup `lpad detect_lostruns`

janosh:
[`detect_lostruns`](https://github.com/materialsproject/fireworks/blob/83c8f8581fc3b617b26141aca1ec06980cbe20f4/fireworks/core/launchpad.py#L1289-L1390) takes 25 min to run on our current `LaunchPad` and looks highly optimizable

1. lines 1351-1361 query each potentially lost FW individually instead of batch fetching (scales O(N) with N = number of fireworks)
2. line 1357 for each non-lost launch, it makes another individual query to check state (nested O(M) queries for launch states where M = number of launches)
3. lines 1378-1388 iterate through all `RUNNING` fireworks and make one query per FW to check if launches are `FIZZLED`/`COMPLETED`

Example: For 10k RUNNING jobs and 1k lost runs, this requires 11,000+ DB queries.

Mongo has had **batch queries and aggregation pipelines** since 3.x to do this faster: (1) Batch fetch all relevant FireWorks with `find({"fw_id": {"$in": fw_ids}})`, (2) Collect all launch IDs then batch fetch their states in one query, and (3) Use a MongoDB aggregation pipeline with `$lookup` to find inconsistent FireWorks server-side instead of N queries client-side. This reduces the operation from ~11,000+ queries to order 10 batch queries plus 1 aggregation, which should give a big speedup and is backwards-compatible

### DIFF
--- a/fireworks/core/firework.py
+++ b/fireworks/core/firework.py
@@ -432,10 +432,10 @@ class Tracker(FWSerializable):
         if os.path.exists(m_file):
             with zopen(m_file, "rt", errors="surrogateescape") as f:
                 for line in reverse_readline(f):
-                    lines.append(line)
+                    lines.append(line.rstrip("\n\r"))
                     if len(lines) == self.nlines:
                         break
-            self.content = "".join(reversed(lines)).rstrip("\n")
+            self.content = "\n".join(reversed(lines))
         return self.content
 
     def to_dict(self):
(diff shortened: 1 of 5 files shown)


janosh:
anecdotally, this PR brings the run time from 25min down to 25sec

```sh
time lpad detect_lostruns
> 0.67s user 0.10s system 3% cpu 24.530 total
```

PR #573, opened 2026-04-03
Bugfix in queue launcher

esoteric-ephemera:
Small bug if the `block_dir` is specified explicitly, `exist_ok` is passed to `os.mkdir` which is only a valid kwarg for `os.makedirs`


computron:
Thanks!

